Issue #943, opened 2014-12-18
Quad Color exports in Board3D does not work anymore 


dcoeurjo:
due to https://github.com/DGtal-team/DGtal/commit/05c5f89628d8806fc4f5468825c11d03d37b11b7 apparently

JacquesOlivierLachaud:
Do you have a corresponding example/test in mind ?

dcoeurjo:
You can have a look to testBoard3d (I don't know if colors are tested). Otherwise, 3dcurvatureviewer with export in dgtaltools

kerautret:
I fix it ;)
In fact the Display3D subclass Board3D was not updated with new Jaco QuadMap.
I update it for the export (in PR #946). For the color,I have adapted the commented createNewQuadList of Display3d with a simple search ok an empty key and an update of the name3d attribute. It works with see DGtalTools PR #194

dcoeurjo:
Hi, I'm not sure to understand. Is the fix to ask the user to add a createNewQuadList  every time he wants to change the color (which is not great IMHO) ?

Why this fix/issue only appears for quad and not other shapes ?

Before your commits in December, this was done in CustomColors3D or setPen (if I remember correctly).

dcoeurjo:
As far as I remember, lists were used in board3d with the following semantic: objects within a list share the same material (to reduce MTL filesize). Now, it may be deprecated since I use now a std::map for material. 

Anyway, what is the semantic you attach to lists in your recent PR?

kerautret:
"Hi, I'm not sure to understand. Is the fix to ask the user to add a createNewQuadList every time he wants to change the color (which is not great IMHO) ?"
=> Hi, 
yes, initially it is inside the drawing factory but it was deactivated in the Jaco PR with map. I can un comment it it should work.
"Why this fix/issue only appears for quad and not other shapes ?"
=> For now just the Quad3D can deal with interaction using map, so for me the global final transformation should be better to be done it in another PR with more semantic discussions between interaction/list.

"Before your commits in December, this was done in CustomColors3D or setPen (if I remember correctly)."
=> It always the case a CustomColors3D implies a new list, It is not the fix itself, it is the case from a long time, once the user use

kerautret:
"As far as I remember, lists were used in board3d with the following semantic: objects within a list share the same material (to reduce MTL filesize). Now, it may be deprecated since I use now a std::map for material."
=> yes it is
[...]

PR #946, opened 2015-01-05
Some cleanings viewer3d

kerautret:
Some cleanings and improve in viewer3d


src/DGtal/io/viewers/Viewer3D.h
```diff
@@ -1028,6 +1028,135 @@ namespace DGtal
...
+     **/
+    void glCreateListPolygonsWired(const std::vector<VectorPolygon>  &aVectPolygon, unsigned int idList);
+
+    
+    /**
+     * Update the container of GLTextureImage object with the given vector of TextureImage. 
+     * @param[in] aVectImage the vector containing 
+     *
+     **/    
+    void glUpdateTextureImages(const std::vector<Viewer3D<Space, KSpace>::TextureImage>  &aVectImage);
```
JacquesOlivierLachaud: Perhaps Create also a type VectorTextureImage

src/DGtal/io/viewers/Viewer3D.ih
```diff
@@ -182,10 +182,18 @@ DGtal::Viewer3D< Space ,KSpace >::updateTextureImage(unsigned int imageIndex, co
   BOOST_CONCEPT_ASSERT(( concepts::CConstImage < TImageType > ));
   assert ( imageIndex< myGSImageList.size());
   typename Viewer3D< Space ,KSpace >::TextureImage &anImage = myGSImageList.at(imageIndex);
-  Display3D< Space, KSpace>::updateBoundingBox(DGtal::Z3i::RealPoint(anImage.point1[0]+xTranslation, anImage.point1[1]+yTranslation, anImage.point1[2]+zTranslation));
-  Display3D< Space, KSpace>::updateBoundingBox(DGtal::Z3i::RealPoint(anImage.point2[0]+xTranslation, anImage.point2[1]+yTranslation, anImage.point2[2]+zTranslation));
-  Display3D< Space, KSpace>::updateBoundingBox(DGtal::Z3i::RealPoint(anImage.point3[0]+xTranslation, anImage.point3[1]+yTranslation, anImage.point3[2]+zTranslation));
-  Display3D< Space, KSpace>::updateBoundingBox(DGtal::Z3i::RealPoint(anImage.point4[0]+xTranslation, anImage.point4[1]+yTranslation, anImage.point4[2]+zTranslation));
+  Display3D< Space, KSpace>::updateBoundingBox(DGtal::Z3i::RealPoint(anImage.point1[0]+xTranslation,
```
JacquesOlivierLachaud: Perhaps define inner types
typedef Display3D< Space, KSpace> Display;
typedef DGtal::Z3i::RealPoint RealPoint;

src/DGtal/io/viewers/Viewer3D.ih
```diff
@@ -623,22 +638,23 @@ DGtal::Viewer3D<Space, KSpace>::drawWithNames()
   // selection is left clic + shift key.
   // JOL: 2014/10/15. This is my addition for interacting with
   // quads. Seems to work well.
-  unsigned int nbListOfPrimitives = Viewer3D<Space, KSpace>::myLineSetList.size() +Viewer3D<Space, KSpace>::myCubeSetList.size()+ Viewer3D<Space, KSpace>::myBallSetList.size();
-  glCallList ( GLuint ( myListToAff + nbListOfPrimitives + 1 ) );
+  //  unsigned int nbListOfPrimitives = Viewer3D<Space, KSpace>::myLineSetList.size() +Viewer3D<Space, KSpace>::myCubeSetList.size()+ Viewer3D<Space, KSpace>::myBallSetList.size();
```
JacquesOlivierLachaud: Dead code ?

src/DGtal/io/Display3D.h
```diff
@@ -424,7 +424,7 @@ namespace DGtal
      * @return the select callback function that match the given \a
      * name, or 0 if none is associated to this name.
      */
-    SelectCallbackFct getSelectCallback3D( DGtal::int32_t name, void*& data ) const;
+    SelectCallbackFct getSelectCallback3D( DGtal::int32_t aName, void*& data ) const;
```
JacquesOlivierLachaud: Need to update the comment above "name" => "aName".
That's why clang fail on travis.

src/DGtal/io/Display3DFactory.h
```diff
@@ -613,7 +613,7 @@ namespace DGtal
      * @param name3d an object storing the "OpenGL" name.
      */
     static void 
-    draw( Display & display, const DGtal::SetName3D& name3d );
+    draw( Display & display, const DGtal::SetName3D& aName3d );
```
JacquesOlivierLachaud: Same comment as above. The comment above should be update "name3d" => "aName3d'.

JacquesOlivierLachaud:
Examples work on my computer. However, the selection (shift+click) of a digital point in some viewer examples (viewer-[2-5]* at least) introduce a seg fault. I think it is because of some former code on set selection and removal.

JacquesOlivierLachaud:
It is much better now to have separate all the codes to generate GL lists.

src/DGtal/io/viewers/Viewer3D.h
kerautret: agree thank, done

src/DGtal/io/viewers/Viewer3D.ih
kerautret: Yes, I add it for Display and also for RealPoint but I change with the one of Viewer3D:
(defined as Space::RealPoint), so I just I switch all  DGtal::Z3i::RealPoint into RealPoint with def from Viewer3d by using:
    typedef typename Display::RealPoint RealPoint;

src/DGtal/io/viewers/Viewer3D.ih
kerautret: yes removed

src/DGtal/io/Display3D.h
kerautret: Yes Thank , I miss it! ;)

src/DGtal/io/Display3DFactory.h
kerautret: done

kerautret:
I see also the set fault, I look it thank

kerautret:
I fix it, it was in fact in getSelectCallback3D, I add a test   if (it_up == mySelectCallBackFcts.begin()){
to check if only one data/func are found.
But by the way, why is the second look look needed ? from line 174 of new Display3D.ih ?

kerautret:
I check the remaining warnings

JacquesOlivierLachaud:
In fact, we need to look for a match twice because we are looking not for an exact, but for a range of values. Hence the find is done with a call to "upper_bound". Then a correct answer is either the returned iterator or the previous value, since ranges have no overlap. (All this is said "as far as I remember").

JacquesOlivierLachaud:
It seems it is no more Travis that is running continued integration. For the error returned by AppVeyor, I guess @dcoeurjo is the best to debunk it.

kerautret:
Ha ok I see :)
yes it looks fine for travis no more any warnings ;)
Thank you

dcoeurjo:
Yeah, sorry for appveyor, I'm debugging it but it's not ready (supposed to be a Travis like for MSwindows VM)

kerautret:
ok so perhaps we can merge?

dcoeurjo:
Appveyor disabled. 
BTW, it may not be related but could you have a look to #943 ? this issue is related to recent refactoring you did on display3D

kerautret:
ok I look it

JacquesOlivierLachaud:
I wait for the answer to David's question before merging:
"As far as I remember, lists were used in board3d with the following semantic: objects within a list share the same material (to reduce MTL filesize). Now, it may be deprecated since I use now a std::map for material.

Anyway, what is the semantic you attach to lists in your recent PR?"

What I did for surfel interaction, is related to assigning a name to surfel(s), but it's true I don't know how this conflicts with lists. I guess we should say that all elements of a list have the same opengl name.

kerautret:
yes I answered, we can perhaps keep only map, the pb that can appears is perhaps the SetName3D( name++ ) which can overwrite an other list generated by CustomColor for instance.

kerautret:
I correct the last remarks with the fix avoiding the user list creation and an color add in testBoard3D with quad.

dcoeurjo:
back on this topic (sorry for being slow on that): beside list usage in Board3D and Viewer3D, can we clarify the list thing and say something like that:
   a list is a collection of objectswhich are actionable

  - Viewer3D:  objects in list have the same OpenGL ID and thus can be "selected" (current situation)
  - Board3D: lists are exported with the notion of OBJ groups meaning that they can be "selected" in a modeler (e.g. blender)
  - Board3Dto2D: no consequence

If we agree, than I should remove my previous usage that objects in a list share the same material (only for perf issues not needed anymore), and remove the fact that changing the color (or mode) create a new list (current behavior).
 Maybe, @kerautret, you should check that being "actionable" is the only usage of list you have in Viewer3D, i don' t know

Would you agree with that?

JacquesOlivierLachaud:
Seems reasonnable to me. I just do not know if an OpenGL list can have different materials. Bertrand, do you know more ?

kerautret:
@dcoeurjo @JacquesOlivierLachaud Yes I confirm the list are not exploited in the spirit of material.
So I am ok for the proposition. Perhaps we merge first this one and then list update ? 
Or else you can PR it on my branch for the review.

JacquesOlivierLachaud:
I suggest to merge and to do another PR afterwards. Ok for you David @dcoeurjo ?

dcoeurjo:
Perfect :)

dcoeurjo:
Done.. Thanks

PR #948, opened 2015-01-12
Minor patch for BezoutVector computation

isivigno:
Bezout vector computation was not correct for some particular cases and led to incorrect createPattern computations. For instance createPattern(Point(0,0),Point(-1,-3)) calls bezoutVector(-3,-1,-1) which returned v=(-1,-4) before this patch, instead of the vector (0,-1), thus missing the lower leaning point (-1,-1).

### DIFF
--- a/src/DGtal/geometry/curves/ArithmeticalDSSFactory.ih
+++ b/src/DGtal/geometry/curves/ArithmeticalDSSFactory.ih
@@ -232,18 +232,34 @@ bezoutVector(const Coordinate& aA,
   //compute the one whose component
   //have a sign equal to the direction
   //vector components
-  if ( (aB >= 0)&&(v[0] < 0) )
-    v += Vector(aB,aA); 
-  else if ( (aB < 0)&&(v[0] >= 0) )
-    v += Vector(aB,aA); 
-  else if ( ( aA >= 0 )&&(v[1] < 0) )
-    v += Vector(aB,aA);
-  else if ( ( aA < 0 )&&(v[1] >= 0) )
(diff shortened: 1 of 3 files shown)


src/DGtal/geometry/curves/ArithmeticalDSSFactory.ih
```diff
@@ -232,18 +232,34 @@ bezoutVector(const Coordinate& aA,
...
+  else if ( ( aA < 0 )&&(v[1] > 0) )
     v += Vector(aB,aA);
+  
   ASSERT( (aA*v[0]-aB*v[1]) == aR );
-  ASSERT( (( aA >= 0 )&&(v[1] >= 0))||(( aA < 0 )&&(v[1] < 0)) ); 
-  ASSERT( (( aB >= 0 )&&(v[0] >= 0))||(( aB < 0 )&&(v[0] < 0)) ); 
+ 
+  // Too restrictive asserts: cases where one v component is null are rejected while they should be accepted. See for instance bezoutVector(-3,-1,-1)
+  //ASSERT( (( aA >= 0 )&&(v[1] >= 0))||(( aA < 0 )&&(v[1] < 0)) ); 
+  //ASSERT( (( aB >= 0 )&&(v[0] >= 0))||(( aB < 0 )&&(v[0] < 0)) );
```
dcoeurjo: could you please cleanup the commented code ?

dcoeurjo:
Thanks, could you please add a bug fix entry in the changelog?

dcoeurjo:
Thanks a lot.. merging.

Issue #964, opened 2015-02-25
greedy-plane-segmentation-ex2 fails at display


kerautret:
I have an error when testing the example greedy-plane-segmentation-ex2:
[ERR] Assertion Error - assertion (myCellEmbedder->isValid()) failed in RealPoint DGtal::Display3D<DGtal::SpaceND<3, int>, DGtal::KhalimskySpaceND<3, int> >::embedK(const typename KSpace::Cell &) const [S = DGtal::SpaceND<3, int>, KS = DGtal::KhalimskySpaceND<3, int>]: /Users/kerautre/EnCours/DGtal/src/DGtal/io/Display3D.ih(1226)

kerautret:
also same with greedy-plane-segmentation-ex3

JacquesOlivierLachaud:
Ok. I look at it.

kerautret:
fixed, https://github.com/DGtal-team/DGtal/pull/967

PR #967, opened 2015-03-01
Fixing Issue 964

JacquesOlivierLachaud:
Fix Issue #964. The problem was related to the update of Display3D, which requires now an instance of KSpace when displaying cells.


kerautret:
yes, perfect thanks, I merge !

PR #968, opened 2015-03-01
fix 931

kerautret:
Changed as given in the assert


JacquesOlivierLachaud:
Is there a problem with travis bot ?

kerautret:
It looks related to qglviewer lib, @dcoeurjo has tested version changed but not sure that ok

dcoeurjo:
Should be ok, you can merge.

PR #971, opened 2015-03-10
fix of Labels iterator

rolanddenis:
Fixing a bug on iterators over Labels that makes begin() iterator equal to end() when the first label is set.

### DIFF
--- a/src/DGtal/base/Labels.ih
+++ b/src/DGtal/base/Labels.ih
@@ -72,7 +72,7 @@ ConstEnumerator( const Word* address, SizeType firstWord )
   else
     {
       myWord = 0;
-      myLabel = 0;
+      myLabel = -1;
     }
 }
 //-----------------------------------------------------------------------------
@@ -132,7 +132,7 @@ operator++()
       if ( myWordLabel == ( __DGTAL_LABELS_NBWORDS
                             * __DGTAL_WORD_NBDIGITS ) )
         {
-          myLabel = 0;
+          myLabel = -1;
           return *this;
         }
       myWord = *myWordAddress++;


JacquesOlivierLachaud:
Fix consistent. Travis ok except for the weird recurrent bug. I merge.

dcoeurjo:
oops too late, I have an issue with this PR: Labels must be unsigned (I don't know why the concept check  has been commented out). I'll comment later..

dcoeurjo:
As default type: label is uint and the -1 initialization is hazardous...

rolanddenis:
Ok. J.-O. suggested initialization with ~0. The fix is coming.

dcoeurjo:
Wait... I also have an issue with ~0: default c++ will interpret this as "int" negate and there could be issues with the implicit cast to words... I'd use std::numeric_limits instead.. @JacquesOlivierLachaud  ?

JacquesOlivierLachaud:
Le 10/03/2015 15:47, David Coeurjolly a écrit :
> Wait... I also have an issue with ~0: default c++ will interpret this as
> "int" negate and there could be issues with the implicit cast to
> words... I'd use std::numeric_limits instead.. @JacquesOlivierLachaud
> <https://github.com/JacquesOlivierLachaud> ?
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/DGtal-team/DGtal/pull/971#issuecomment-78068535>.
> 
For me, ~ is a bitwise operator and does not care about integer sign.
Hence, it flips all 0 to 1 without any check.

rolanddenis:
Ho, I thinks std::numeric_limits was from the C++11 norm ... apparently not. I agree that it is the best solution.

rolanddenis:
I have updated PR #972 .

dcoeurjo:
Thanks @rolanddenis !!
Sorry about that, I am have been a bit paranoiac with this implicit cast stuff...

( @JacquesOlivierLachaud  ok for ```~``` but I have the feeling that ```~static_cast<Label>(0)```would have been more secure.. again, maybe I'm crazy with that...)

JacquesOlivierLachaud:
But numeric_limits is fine also !

PR #972, opened 2015-03-10
Using ~0 instead of -1 as magic number for end().

rolanddenis:
Using -1 as magic number for unsigned int is implementation dependant.
~0 is a better choice and is type-size independant.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -62,6 +62,8 @@
  Lachaud, [#926](https://github.com/DGtal-team/DGtal/pull/926))
 
 - *Base Package*
+ - Fix bug with Labels iterator when first index is set (Roland Denis,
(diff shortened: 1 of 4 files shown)


dcoeurjo:
Thanks. It was not in your PR but could you please uncomment the CONCEPT_ASSERT in Labels.h  (and check what was wrong) ?

dcoeurjo:
(please have a look to my comment on the previous PR)

rolanddenis:
Concerning the CONCEPT_ASSERT, the concept name was wrong. I corrected it (to CUnsignedNumber) and added CIntegralNumber concept check. I also added a test block in testLabels.cpp related to the iterator bug.
Could you please wait to merge this PR: I have another bug with LabelledMap that I wish to include in this PR.

dcoeurjo:
Great. thanks. 
Merging

PR #973, opened 2015-03-11
Fixing LabelledMap copy constructor and copy operator

rolanddenis:
There was several issues with copy constructor and copy operator of LabelledMap. The test file has been updated to check this issues.

The documentation has been corrected (typo in the first block capacity).


dcoeurjo:
Great! thanks

PR #981, opened 2015-03-17
Update of  FindQGLViewer.cmake script for better discovery in Mac OSX

dcoeurjo:
,


kerautret:
I check it

kerautret:
Looks fine for me at least on my mac (with qglviewer home installation).
By the way have tested with qt5 ? @naubry has some issues but perhaps it solve the pb or not related so I merge ;)

PR #982, opened 2015-03-23
Fixing "uninitialized variable" warning with EuclideanShapeDecorator

jlevallois:
Fixing warnings + ASSERTS ...

:octocat:

### DIFF
--- a/src/DGtal/shapes/DigitalShapesDecorator.h
+++ b/src/DGtal/shapes/DigitalShapesDecorator.h
@@ -164,6 +164,7 @@ namespace DGtal
       if( !isValid() )
       {
         trace.error() << "Operation unvalid. Maybe you don't set a ShapeA object." << std::endl;
+        ASSERT(( isValid() ));
         return;
       }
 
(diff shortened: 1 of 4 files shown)


src/DGtal/shapes/DigitalShapesDecorator.h
```diff
@@ -164,6 +164,7 @@ namespace DGtal
       if( !isValid() )
       {
         trace.error() << "Operation unvalid. Maybe you don't set a ShapeA object." << std::endl;
+        ASSERT(( isValid() ));
```
dcoeurjo: kittens are dying... :crying_cat_face: 

Please consider DGtal::FATAL_ERROR or FATAL_ERROR_MSG to have a non-static assert even in release mode

jlevallois: It's wasn't in DGtal coding guide to never assert/crash in release mode ?

dcoeurjo:
Dudes don't need guidelines... ;) 

thanks for the edits.. Waiting for the travis

PR #983, opened 2015-03-25
Adding support of Qt5 for QGLViewer that was compiled with it

naubry:
Hi,

I'm trying to add the support of Qt5 in order to use QGLViewer that was compiled with it. I'm not sure that is the best way to integrate it but I made the maximum to keep both functionnalities (The possibility to use Qt4 and/or Qt5).

I changed two files : CheckDGtalOptionalDependencies.cmake and Viewer3D.h.

In the Viewer3D.h file, I only changed one line and I think it's working on both Qt4 and Qt5 without a conditionnal test...

I remain at your disposal for any questions or modifications.


kerautret:
Great, I look it, and by the way have you check if DGtalTools is also fine with QT5 ?

naubry:
No, sorry I haven't verified if DGtalTools works with Qt5 and my modification.

naubry:
I made some modifications in DGtal and DGtaltools and I have compiled 99% of DGtalTools for the moment..

kerautret:
I just tested on my config and from the construction i obtain:
----------------
 CMake Error at cmake/CheckDGtalOptionalDependencies.cmake:340 (find_package):
   By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
   asked CMake to find a package configuration file provided by "Qt5", but
   CMake did not find one.

   Could not find a package configuration file provided by "Qt5" with any of
   the following names:

     Qt5Config.cmake
     qt5-config.cmake

   Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
   to a directory containing one of the above files.  If "Qt5" provides a
   separate development package or SDK, be sure it has been installed.
 Call Stack (most recent call first):
   CMakeLists.txt:40 (INCLUDE)

naubry:
You must specify by hand the path where cmake can find qmake and where are the Qt5config files.
Variables are : QT_QMAKE_EXECUTABLE and Qt5_DIR

For me for instance, qmake is here : /Users/naubry/Qt/5.4/clang_64/bin/qmake
and the Qt5config files : /Users/naubry/Qt/5.4/clang_64/lib/cmake/Qt5

I have used cmake-gui to add these variables, it's more convenient :)

kerautret:
My last message was from my current build, so I tried with an empty one I  have no error message but with your change it doesn't consider QT 4 so when QGLviewer is activated with ccmake construct as all as if it was not present. ;(

cmake/CheckDGtalOptionalDependencies.cmake
```diff
@@ -21,7 +21,7 @@ OPTION(WITH_MAGICK "With GraphicsMagick++." OFF)
 OPTION(WITH_ITK "With Insight Toolkit ITK." OFF)
 OPTION(WITH_CAIRO "With CairoGraphics." OFF)
 OPTION(WITH_HDF5 "With HDF5." OFF)
-OPTION(WITH_QGLVIEWER "With LibQGLViewer for 3D visualization (Qt required)." OFF)
+OPTION(WITH_QGLVIEWER_QT "With LibQGLViewer for 3D visualization (Qt required)." OFF)
```
dcoeurjo: Please keep the WITH_QGLVIEWER flag, this flag is used at many places in the code.

cmake/CheckDGtalOptionalDependencies.cmake
```diff
@@ -335,21 +335,53 @@ ENDIF(WITH_HDF5)
 # -----------------------------------------------------------------------------
 # Look for Qt (needed by libqglviewer visualization).
 # -----------------------------------------------------------------------------
+set(QT5_FOUND_DGTAL 0)
+if (WITH_QGLVIEWER)
+  find_package(Qt5 COMPONENTS Widgets OpenGL Xml REQUIRED)
```
dcoeurjo: the issue comes from here and the REQUIRED

dcoeurjo:
I would suggest
* to keep QT4 as a (default) dependency (with ```REQUIRED```)  when WITH_LIBQGLVIEWER is on 
* to create an additional WITH_QGLVIEWER_QT5 flag for advanced users (default "off") such that if the flag is set to true, then the FIND will search for QT5 instead of QT4

dcoeurjo:
@naubry does it make sense for you ?

naubry:
That makes sense to me. I have already started working in this direction last night.

naubry:
I have added an option to activate Qt5. 
DGtal lib, examples and tests compile and run on both system Linux and OSX.
I have also modified DGtalTools with success. So, I can make a PR for this modifications soon.

naubry:
Sorry, I made a mistake when I merged my new code with this branch...

kerautret:
Looks fine for me, tested with default QT4 and QT5 (we need to specify the qt5 dir:Qt5_DIR=/Users/kerautre/Qt/5.4/clang_64/lib/cmake/Qt5/) .
It looks also ok in DGTalTools (by the way Nicolas you can PR in main DGtalTools ? )
@dcoeurjo for DGtalTools I needed to apply :--DQt5_DIR=/Users/kerautre/Qt/5.4/clang_64/lib/cmake/Qt5/
I don't remember how it can be transmitted automatically in DGtaTools ?

dcoeurjo:
Thanks @naubry. There is still something that puzzles me. The QT includes: 

     #ifdef WITH_QT5
        #include <QApplication>
    #else
       #include <QtGui/qapplication.h> 
    #endif

in the examples seem to be duplicated since there are already included in Viewer3D. I know some of them were already there before your QT5 thing but maybe we should cleanup this up.

dcoeurjo:
@kerautret See the  the ```DGtalConfig.cmake.in``` line 144, maybe just add a ```set(QT5_DIR @QT5_DIR@)``` ?

@kerautret can you make sure that in the doc, we only require to include "Viewer3D.h" and not the Qt stuff when using the viewer ?  (Viewer3D should be in charge to include all qt deps)

kerautret:
thanks @dcoeurjo yes I confirm it works (just lower case for the name).
@naubry I PR on your branch

kerautret:
I look it for the doc

kerautret:
@dcoeurjo I confirm the doc don''t refer to any QT include...

dcoeurjo:
Perfect @naubry  thanks a lot.. 
Congrats for your first PR !*
(I'll add you to the AUTHOR list)

PR #985, opened 2015-03-25
Change export color mode in OFF Mesh 

kerautret:
It was not natural to have the default option set to false when dealing with Mesh storing colors.
Now, when exporting mesh it first tries to export colors if available. (you can always force to not do it by changing the default option).


dcoeurjo:
Ok thanks, merging

PR #989, opened 2015-04-08
Fix basic domain sub sampler

kerautret:
BasicDomainSubSampler can now handle non 0 origin point. This update also correct the search of point which are outside the source domain.
It also fix the issue #210 of DGTalTools.


kerautret:
@dcoeurjo it also fix another issue of meshViewer (see PR of DGTalTools #211)

dcoeurjo:
ok thanks 
merging.. we'll see later for travis status

PR #991, opened 2015-04-09
fix warning std::abs

kerautret:
Small PR just to remove warning appeared from last OS X update


rolanddenis:
Actually, it is more than a simple warning: if both `abs` and `std::abs` are in scope, and without explicit use of `std::abs` (or of the directive `using std::abs;`), the compiler will choose `abs` that works on integers.

For example, the following code:
```C++
#include <iostream>
#include <cstdlib>
#include <cmath>

int main()
{
    double v = 2.5;
    std::cout << "abs(2.5) = " << abs(v) << std::endl;
    std::cout << "std::abs(2.5) = " << std::abs(v) << std::endl;
    return 0;
}
```
compile silently with `g++` (v4.9.1 on Linux), even with `-Wall -Wextra`, and with a warning with `clang++` (v3.5.0):
```
test.cpp:8:35: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
    std::cout << "abs(2.5) = " << abs(v) << std::endl;
                                  ^
test.cpp:8:35: note: use function 'std::abs' instead
    std::cout << "abs(2.5) = " << abs(v) << std::endl;
                                  ^~~
                                  std::abs
1 warning generated.
```

 but both binaries output is:
```
abs(2.5) = 2
std::abs(2.5) = 2.5
```

I've made a fast search through the DGtal source and it seems to be the only file where abs (alone) is used on floats.

kerautret:
yes,  I also detects strange behavior with abs in my dev prog ;)

dcoeurjo:
Thanks @kerautret for the PR and @rolanddenis for the explanation :)

dcoeurjo:
@kerautret could you please merge the master to this branch?

kerautret:
ok

kerautret:
@dcoeurjo done

dcoeurjo:
merging.

PR #994, opened 2015-04-15
Fix bug of 3dCurvatureViewer and 3dCurvatureViewerNoise

kerautret:
The bounds were never good in the scanner (lower dim).
Perhaps @JacquesOlivierLachaud  can confirm since I back port it from ImaGene ;)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -131,6 +131,9 @@
     the search of point which are outside the source domain (it is now checked in testBasicPointFunctors).
     (Bertrand Kerautret, [989](https://github.com/DGtal-team/DGtal/pull/989)).
 
+- *Topology  Package*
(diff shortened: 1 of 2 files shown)


kerautret:
It works with @akrah computer, perhaps just wait the Markus test ;)

JacquesOlivierLachaud:
Does the fix hide another bug in KSpace::lowerBound and upperBound ?

kerautret:
Yes perhaps since I changed the KSpace::lowerBound  into KSpace::lowerCell but it could be also the KSpace::uNext( p, dir_low_uid, dir_up_uid ) which continue when cell of lower dim are given ?

dcoeurjo:
Thanks, merging

PR #996, opened 2015-04-16
Adding support of empty HyperRectDomain.

rolanddenis:
This is not a very useful feature but I need to be able to run DGtal (in `DEBUG` mode) using empty domains (coming from a view on an image, that behaves like a `CImage` but with domain equal to the bounding box of the original image, which may be empty).

A `HyperRectDomain` is empty iff `lowerBound = upperBound + diagonal(1)`. This arbitrary definition is chosen so that `begin() == end()` on empty domain.

There is no change in the code, except that:
- the default constructor of `HyperRectDomain` now creates an empty domain, instead of the domain containing the origin,
- there is a new method, `isEmpty`, that returns `true` if the domain is empty.

The other modifications are about the assertions that failed on empty domain.
In addition, there was a bug in those assertions when comparing bounds of the domain or when testing if a point lies inside the domain (when `isInside` was not used, which behaves properly), because of the use of the `<=` operator. The comparison operators are inherited from the container of `PointVector`, which used the lexicographical order for the default container. 
Consequently, the following code was able to run without error in `DEBUG` mode:
```C++
typedef SpaceND<2> Space;
typedef Space::Point Point;
HyperRectDomain<Space> domain( Point::diagonal(0), Point::diagonal(10) );
const Point point(1, -1);
auto it = domain.subRange( 0, 1, Point::diagonal(0) ).begin( point );
```
because `(1,-1)` is lexicographically greater that `(0,0)`. 

The other modifications are about:
- adding messages to the assertions,
- cleaning (deleting tabulations and reindenting) the code,
- deleting useless member `myCurrentPos` in HyperRectDomain iterators,
- documenting (warning) about the use of lexicographical order in comparison operators of `PointVector`,
- adding tests on empty domains.


dcoeurjo:
Thanks @rolanddenis, this makes sense. Maybe we should add a "@advanced" note in the ```moduleSpacePointVectorDomain.dox``` documentation page ?

rolanddenis:
I've added a small section about empty domains and made some minor modifications on the layout of `moduleSpacePointVectorDomain.dox`.

dcoeurjo:
thanks a lot that's perfect

PR #1001, opened 2015-04-27
Adding of a little more genericity in ImageHelper.h

rolanddenis:
Changes are:
- enabling `imageFromImage` for two different images type with `CConstImage` concept check on source image,
- `CConstImage` concept check and const-correctness in `ImageToConstantFunctor`,
- checking `CConstImage` instead of `CImage` in `findAndGetValue`.


dcoeurjo:
thanks a lot.. merging

PR #1009, opened 2015-05-09
SVG cleanup in DEC package

dcoeurjo:


dcoeurjo:
self merging (to review DEC packages)

PR #1012, opened 2015-05-18
Removing warnings using last MacOS clang (6.1.0)

dcoeurjo:


kerautret:
great no more warnings in main and tests, thank merging ;)

kerautret:
you should first update from master ;)

PR #1015, opened 2015-05-26
add potential use of int (ASCII) mode in PGM writer

kerautret:
To use not only 255 max levels in PGM/PGM3D writer..


kerautret:
not ready the concept_check is not fine,I try another solution with template partial specialization

PR #1016, opened 2015-05-28
add color in MeshWriter to export obj file

kerautret:
Small edits to have color when exporting mesh in OBJ format (was already done in OFF but not in OBJ)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -94,6 +94,8 @@
  Aubry, [#983](https://github.com/DGtal-team/DGtal/pull/983))
  - Board2D now supports quadratic Bezier curve drawing. (Tristan Roussillon,
   [#1002](https://github.com/DGtal-team/DGtal/pull/1002))
+ - MeshWriter class can now export OBJ file including colors. (Bertrand Kerautret,
+  [#1016](https://github.com/DGtal-team/DGtal/pull/1016))
 
 - *Kernel Package*
   - HyperRectDomain can now be empty (lowerBound == upperBound + diagonal(1)). Warning about the use 
--- a/src/DGtal/io/writers/MeshWriter.h
+++ b/src/DGtal/io/writers/MeshWriter.h
@@ -97,7 +97,7 @@ namespace DGtal
 
 
     /** 
-     * Export a Mesh towards a OBJ format.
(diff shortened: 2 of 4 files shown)


dcoeurjo:
For Obj export of meshes,  is   ```board3D<< aMesh; board3D.saveOBJ()``` equivalent to exportObJ ?
I don't know if we need to duplicate the code for similar feature.

kerautret:
perhaps quite similar, but for me I am not sure that a first user will have the reflex to use Board just to export a mesh object ? As you want but for me it is more logic if Board call MeshWriter to export than the inverse. But I can call the board writer

kerautret:
... in the current export2OBJ_colors functions...

dcoeurjo:
The export code in Board is also shared by Viewer3d. 
The difference I see is could be that board exports the mesh as a set of triangles whereas your code exports it as a triangular surface.. 
Could you please confirm me ?

kerautret:
Yes I confirm, it takes all the faces which are not necessary a triangles.

dcoeurjo:
Actually board can also export polygonal faces but I meant that for triangular meshes, vertices are duplicated in board which may not be the case in your export. Do you agree ?

kerautret:
Ha ok, yes vertices are only given once with also materials with colors.

dcoeurjo:
@kerautret the testColor.obj does not show up colors on meshlab. have you used something specific ? Please check for instance material syntax we use in Board3D

kerautret:
I think that it MeshLab which by default apply a color rendering by vertex, but here it is by face ;)
I don't know why it does't check it by default, but if you check the option in render menu -> Color -> Per Face it should be fine in theory...

dcoeurjo:
Maybe be I'm wrong but after checking the wikipedia description of OBJ, per-vertex-color seems to be more standard than face color

kerautret:
ok, however since it is given for mesh as face color it is logic to do it.
We can export it perhaps later when Mesh will have color on vertex?

dcoeurjo:
as you whish

kerautret:
ok for me for merging and I will add an issue dealing with color on vertex with obj export

dcoeurjo:
done

dcoeurjo:
thanks

PR #1019, opened 2015-06-09
C++ unit test framework with Catch

dcoeurjo:
Enables catch unit test  system (https://github.com/philsquared/Catch) in DGtal.
The idea is to simplify unit test design in DGtal with a "single-header" but complete framework.

For instance, a ```REQUIRE( a == b);``` in a test case section checks that ```a``` equals ```b```. If the test fails, the values of ```a``` and ```b``` are traced out with neat messages.

Basically it would replace what we usually do:

     trace.info() << "Checking if "<<a<< " equals "<<b<<std::endl;
     nbok += (a == b) ? 1 : 0; 
     nb++;
     trace.info() << "(" << nbok << "/" << nb << ") "
           	       << "(a == b) returns true" << std::endl;

Catch also allows test case filtering and timing...

Please have a look to ```testPointVector-catch.cpp``` which is a catch version of ```testPointVector.cpp```


dcoeurjo:
BTW, rewriting the point vector unit test allows me to detect an issue in PointVector::getNormalized() :smile: 



:octocat:

dcoeurjo:
Run ```testPointVector-catch -h``` to get the (automatic) command line arguments.
E.g.
```testPointVector-catch -s -d``` shows you the successful tests with timings

File: tests/base/testCatch.cpp [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+ *
+ *
+ * @date 2015/08/06
+ *
+ * This file is part of the DGtal library
+ */
+
+/**
+ * Description of test_trace' <p>
+ * Aim: simple test of \ref MeasureOfStraighLines
```
JacquesOlivierLachaud: I guess you mean "catch functionnalities" here.

tests/kernel/testPointVector-catch.cpp
```diff
@@ -0,0 +1,131 @@
...
+ *
+ *
+ * @date 2015/06/06
+ *
+ * This file is part of the DGtal library
+ */
+
+/**
+ * Description of test_trace' <p>
+ * Aim: simple test of \ref MeasureOfStraighLines
```
JacquesOlivierLachaud: Same as above

dcoeurjo:
thx. done.

JacquesOlivierLachaud:
Sinon, dans catch, ils spécifient qu'il faut mettre le #define and
#include catch que dans un seul fichier, puis mettre les tests dans
d'autres fichiers. cf
https://github.com/philsquared/Catch/blob/master/docs/slow-compiles.md

On fait comme ça, où on remettra #define et #include catch dans chacun
des tests à venir ?

dcoeurjo:
We could easily create a ```DGtalCatch.h```with  the #define and #include "catch.h" and force all catch-enabled test in the template to include this file. I'll try this.

dcoeurjo:
Done.

JacquesOlivierLachaud:
According to the pointed doc, it seems the define main+include sould be in one *source* code only, not a header. Then, we only do the include at other places. Have I misunderstood something ?

dcoeurjo:
Yes but we don't want to collect all unit tests into a unique executable...

dcoeurjo:
I don't see any better solution..

JacquesOlivierLachaud:
I was just wondering if the catch source should be compiled within DGtal library (then we can still split tests). The problem is that we don't want to add sources to the library (in fact, we would like to get rid of all cpp, no ?). Perhaps it would be better to have a DGtalTest library. What do you think ?

dcoeurjo:
The problem is more related to the magical ```#define CATCH_CONFIG_MAIN```  it has to be placed before the ```#include "catch.hpp"``` and generates a single ```main```.

Unless  we want to gather all unit tests into a single exe (and using the filter features to focus on the API we want to test), I think there is nothing we can do (beside the #pragma thing that could speed up the build on modern compilers).

JacquesOlivierLachaud:
Ok. We just have to watch the compilation time when there will be many "catchy" tests.
I merge.

Issue #960, opened 2015-02-20
missing examples in "examples" section of doxygen doc


kerautret:
Some examples are not visible in http://liris.cnrs.fr/dgtal/doc/nightly/examples.html
In particular the ones of examples/geometry/curves/
(exampleArithDSS3d, exampleArithmeticalDSL, exampleArithmeticalDSSComputer, exampleDSLSubsegment, exampleFrechetShortcut, exampleGridCurve2d, exampleGridCurve3d, exampleGridCurve3d-2, exampleStabbingCircleComputer, exampleStabbingLineComputer)

You can just add in moduleArithDSSReco.dox something like this example:
/**
This example shows the recognition of a simple standard digital
plane with the Chord algorithm (class
ChordGenericStandardPlaneComputer ). Green points belong to the naive
plane. Grey points show farther points that also belong to the
strip.

@see \ref modulePlaneRecognition_sec26

@verbatim
$ ./examples/io/viewers/viewer3D-7bis-stdplane
@endverbatim

@image html coba-stdplane.png "Piece of standard plane."
@image latex coba-stdplane.png "Piece of standard plane." width=5cm

\example io/viewers/viewer3D-7bis-stdplane.cpp
*/

troussil:
done, see  #1026

dcoeurjo:
Fixed.

PR #1026, opened 2015-07-16
Examples of geometry/curves

troussil:
In this set of commits, I added descriptions of example files (of geometry/curves) in doxygen files so that they are now in the list of examples (see issue #960).


dcoeurjo:
Perfect,thank you.

PR #1029, opened 2015-07-17
Read-write image view from any random-access iterator, sub-domain view and associated tools.

rolanddenis:
The main feature of this PR is to add a read-write image view that works with any random-access iterator and a definition domain. In addition, it is possible to restrict the view to a given sub-domain.
This view also provides efficient iterators that can return the associated point, making it a more efficient way than using a domain iterator and operator()/setValue methods.

Thanks to some helpers, it is easily usable with any CImage model if its begin() method returns a random-access iterator (like a ImageContainerBySTLVector).

Unfortunately, some of the added classes needs C++11 features in order to simplify their usage.

The added files are:
- src/DGtal/kernel/domains/Linearizer.h provides linearization of point inside a given HyperRectDomain (with storage order specifier) and the reverse process (index to point).
- src/DGtal/base/IteratorCompletion.h provides reverse iterators, ranges and associated typedefs for any class that have a minimal set of methods (begin, end, ...) returning iterators.
- src/DGtal/images/ArrayImageIterator.h provides efficient iterator over a (sub-)domain for any image-like class that provides a dereference method that takes a point and a linearized index. This file needs C++11.
- src/DGtal/images/ArrayImageAdapter.h (described above) needs C++11.
- tests/images/testArrayImageAdapter.cpp that tests ArrayImageAdapter.


rolanddenis:
Two remarks:
- I recently swap template order of ArrayImageView (for some strange reason) but I now realize that it can be confusing that this order is not the same as for others image models (Domain first).
- The helpers makeArrayImageViewFromIterator and makeArrayImageViewFromImage should maybe be in a specific namespace, like DGtal::helpers ?

src/DGtal/images/ArrayImageAdapter.h
```diff
@@ -0,0 +1,564 @@
...
+  class ArrayImageAdapter< TArrayIterator, HyperRectDomain<TSpace> >
+      : public IteratorCompletion< ArrayImageAdapter< TArrayIterator, HyperRectDomain<TSpace> > >
+    {
+
+    // Checks Random-access iterator concept on TArrayIterator
+    BOOST_CONCEPT_ASSERT( (boost_concepts::RandomAccessTraversalConcept<TArrayIterator>) );
+
+    public:
+      // Typedefs
+      using Self = ArrayImageAdapter<TArrayIterator, HyperRectDomain<TSpace> >;
```
dcoeurjo: just a quick question, is ```using``` syntax equivalent to ```typedef``` ?

dcoeurjo:
Sorry for the delay but I'm reviewing the PR.

Let's start with Linearizer: There  is an inner class in ImageContainerBySTLVector (linearizer) that seems to do the same thing..
Can you please have a look  ? Maybe you're class is more generic.

dcoeurjo:
For the C++11 requirement, this is a real open-question. At this point, C++11 is not required and only few feature need it (with proper #ifdef protections). I'ven't yet check the code but which c++11 feature do you need in your image adapter class?

src/DGtal/images/ArrayImageAdapter.h
dcoeurjo: if it is, please switch to ```typedef```, all inner types in dgtal  use this syntax (with associated doxygen doc).

rolanddenis:
Yes, ```ImageContainerBySTLVector``` has already a linearizer but this new class has more features:
- linearization (point to index) and de-linearization (index to point) that is needed is the other classes;
- row-major and col-major storage order compatible (maybe not very useful ...);
- no loops, whatever the dimension ("meta-programming").

rolanddenis:
Concerning ```C++11```, it is used in ```ArrayImageIterator.h``` (include by ```ArrayImageAdaper.h```) with ```decltype```, ```enable_if``` and ```is_convertible```. Since the last two are easily replaceable by some boost features, the first one is more tricky and the easiest way would be to add an another template parameter to ```ArrayImageIterator```, which parameter must be different if a mutable or a constant iterator is needed (ie, the syntax is more complex, but not a lot more).

In addition (but I understand that it is not a so easy debate), latest versions of ```gcc``` ( >= 5.0 ) and ```clang``` ( >= 3.6 ) now enables ```C++11``` by default.

src/DGtal/images/ArrayImageAdapter.h
rolanddenis: If we consider that this class stays in ```C++11```, ```using``` is the preferred syntax for typedefs, with a more clear syntax and some more features like templated aliases. 
But that does not prevent them to be documented ;)

rolanddenis:
```decltype``` is also used in ```ArrayImageAdapter.h```, in the ```makeArrayImageAdapterFromImage``` helper to automatically choose the right iterator mutability given the image mutability ( only constant iterators if the image is given by const-ref).

**Comment by dcoeurjo** [comment]
For Linearizer: maybe we could do a separate PR with only this feature (independent of the C++11 discussion). If this linearizer provides faster point-to-index then update the ImageContainerBySTLVector to use this one (and cleanup the previous version). What would you think ?

**Comment by rolanddenis** [comment]
Ok, I'll do a separate PR.
What about ```IteratorCompletion.h``` ?

**Comment by rolanddenis** [comment]
Done: #1039

**Comment by dcoeurjo** [comment]
I'll have a look (to your new PR and IteratorStuff ;)). Thanks

**Comment by dcoeurjo** [comment]
For IteratorCompletion, let me check with @troussil (he wrote plenty of IteratorAdpaters/Wrappers). E.g. there are classes that construct ranges from forward/bidir/../ iterators.

At some point, we need to have a complete documentation (technical one) which describes all adapters there exist in DGtal.. I'm a bit lost here.
 Have you checked that nothing already does the same thing?

**Comment by rolanddenis** [comment]
Yes I've already checked and I see nothing like this but I may have miss something (I'm a bit lost too :) ).

**Comment by troussil** [comment]
There exists ConstRangeAdapter that builds a bidirectionnal range onto a pair of iterators. IteratorCompletion is however different and new in DGtal. It adds to a class that provides basic iterator services, extra iterator services (reverse iterators from bidirectionnal iterators, etc.). I guess it is used to transform an array to a model of CImage.

**Comment by rolanddenis** [comment]
Yes !

**Comment by rolanddenis** [comment]
I have added example files for `ArrayImageAdapter` and `IteratorCompletion`, and modified Images and Iterators documentation. Please have a look ;)

File: src/DGtal/images/ArrayImageAdapter.h [hunk 2]
```diff
@@ -0,0 +1,544 @@
...
+    public:
+      // Typedefs
+      using Self = ArrayImageAdapter<TArrayIterator, HyperRectDomain<TSpace> >;           ///< Self type.
+      using ArrayIterator   = TArrayIterator;                                             ///< The given random-access iterator's type.
+      using Value           = typename std::iterator_traits<ArrayIterator>::value_type;   ///< The value type stored in the image.
+      using Reference       = typename std::iterator_traits<ArrayIterator>::reference;    ///< Mutable reference type.
+      using ConstReference  = const Reference;                                            ///< Constant reference type.
+
+      // DGtal typedefs and constant
+      using Domain    = HyperRectDomain<TSpace>;      ///< Domain type.
```
dcoeurjo: I'll have to check if oxygen is ok with this using syntax ;)

src/DGtal/images/ArrayImageAdapter.h
```diff
@@ -0,0 +1,544 @@
...
+
+      /** Reads a value given a point lying inside the full domain.
+       *
+       * @param[in] aPoint  The point.
+       * @return a constant value.
+       */
+      inline
+      Value getValue( Point const& aPoint ) const
+        {
+          BOOST_ASSERT_MSG(
```
dcoeurjo: there is also a ```ASSERT_MSG``` in DGtal (Assert.h)

src/DGtal/images/ArrayImageAdapter.h
```diff
@@ -0,0 +1,544 @@
...
+
+      /** Sets a value given a point lying inside the full domain.
+       *
+       * @param[in] aPoint  The point.
+       * @param[in] aValue  The value.
+       */
+      inline
+      void setValue( Point const& aPoint, Value aValue )
+        {
+          BOOST_ASSERT_MSG(
```
dcoeurjo: same here

src/DGtal/images/ArrayImageIterator.h
```diff
@@ -0,0 +1,294 @@
...
+   */
+  template <
+    typename TIterableClass
+  >
+  class ArrayImageIterator
+    : public boost::iterator_facade <
+        ArrayImageIterator<TIterableClass>,
+        typename TIterableClass::Value,
+        std::random_access_iterator_tag,
+        decltype( ((TIterableClass*)nullptr)->dereference( TIterableClass::Point::diagonal(0), typename TIterableClass::Point::Coordinate(0) ) )
```
dcoeurjo: ;) learning something here

src/DGtal/images/ArrayImageAdapter.h
dcoeurjo: ok doxygen displays the using but not the type documentation.. typedefs seem to be better..

src/DGtal/images/ArrayImageAdapter.h
```diff
@@ -0,0 +1,544 @@
...
+namespace DGtal
+{
+
+  /////////////////////////////////////////////////////////////////////////////
+  /**
+   * @brief Aim: Image adapter for generic arrays with sub-domain view capability.
+   *
+   * Description of template class 'ArrayImageAdapter' <p>
+   * It is an empty class that is specialized for HyperRectDomain (see corresponding documentation page).
+   *
```
dcoeurjo: Please add a link to the specialized class doxygen documentation (DGtal::ArrayImageAdapter< TArrayIterator, HyperRectDomain< TSpace > >). If you look for ArrayImageAdapter in the doc you get the doc of this empty class.

src/DGtal/images/ArrayImageAdapter.h
rolanddenis: What do you mean by no "type documentation" ? On my side, I do not see the difference with `typedef`.

src/DGtal/images/doc/moduleImages.dox
```diff
@@ -394,7 +398,68 @@ The values associated to writing the points are adapted
 
 The use is the same that for ConstImageAdapter.
 
- \section dgtalImagesFunctions Useful classes and functions
+\subsection arrayimageadat ArrayImageAdapter (C++11 only)
```
dcoeurjo: remove the C+11 only (not needed anymore).
and add yourself to the author list;)

tests/images/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ if( WITH_HDF5 )
   testImageFactoryFromHDF5)
 endif( WITH_HDF5 )
 
+if( WITH_C11 )
```
dcoeurjo: cleanup C++11 test here

tests/images/testArrayImageAdapter.cpp
```diff
@@ -0,0 +1,298 @@
...
+/**
+ * @author Roland Denis (\c roland.denis@univ-smb.fr )
+ * LAboratory of MAthematics - LAMA (CNRS, UMR 5127), University of Savoie, France
+ *
+ * @date 2015/07/16
+ *
+ * This file is part of the DGtal library
+ */
+
+#if __cplusplus < 201103L
```
dcoeurjo: the same here

dcoeurjo:
Thanks again for this nice PR ;)

src/DGtal/images/ArrayImageAdapter.h
dcoeurjo: My bad... I thought that doxygen append the type documentation ```//<``` in the public type section of the resulting HTML file. But it is not the case for ```typedefs``` as well

src/DGtal/images/ArrayImageIterator.h
rolanddenis: That was why I need `C++11` on this class ;)

src/DGtal/images/ArrayImageAdapter.h
rolanddenis: I know but I don't find out how to link to a template specialized class... any idea ?

dcoeurjo: At least we could add a relative link to the HTML page.  but the ```DGtal::ArrayImageAdapter< TArrayIterator, HyperRectDomain< TSpace > >``` symbol is known in doxygen so it should be accessible

rolanddenis: I agree but I don't understand how to link to a templated class: Doxygen do not recognize it if I write `DGtal::ArrayImageAdapter< TArrayIterator, HyperRectDomain< TSpace > >`... 

The only way I see is to tell Doxygen to ignore the generic definition so that he only sees the specialized definition.

dcoeurjo:
Great. thanks.
merging

PR #1034, opened 2015-08-03
Warning fixes again

dcoeurjo:
I'm reopening this branch thanks to recent edits proposed by @rolanddenis  (thanks again @rolanddenis).

Here you have @rolanddenis 's original post:

#### With g++ 4.9.2 (Debian 3.16.7)
Using following options:
BUILD_EXAMPLES
BUILD_SHARED_LIBS
BUILD_TESTINGS
WITH_C11
WITH_CGAL
WITH_EIGEN
WITH_GMP
WITH_HDF5
WITH_MAGICK
WITH_OPENMP
WITH_QGLVIEWER
and compile flags `-Wall -Wextra`, the remaining warnings are about the following points:
- Ignoring warning 'assuming signed overflow does not occur when assuming that (X + c) >= X is always true' in src/DGtal/geometry/volumes/distance/ExactPredicateLpPowerSeparableMetric.ih:294 (-Wstrict-overflow activated by -Wall). It is triggered because of compiler optimization. Should we add a compiler exception for this function ?
- Ignoring "array subscript is above array bounds" in DGtal/io/boards/Board3D.ih:181 (-Warray-bounds activated by -Wall). In debug mode, there is an assert to verify that the index is in the bounds. In release, because of this assert is ignored, the compiler complained about bounds checking. We should mayby throw an exception ?
- Ignoring "‘aValue’ may be used uninitialized in this function" in src/DGtal/images/TiledImage.h:710 (-Wmayby-uninitialized activated by -Wall). There is no trivial solution to initialize aValue. Alternatively, we should get the read result on line 728 and throw an exception if failed.

#### With clang++ 3.5.0 (Debian 3.16.7)
Using following options:
BUILD_EXAMPLES
BUILD_SHARED_LIBS
BUILD_TESTINGS
WITH_C11
WITH_CGAL
WITH_EIGEN
WITH_GMP
WITH_HDF5
WITH_MAGICK
~~WITH_OPENMP~~
WITH_QGLVIEWER
and compile flags `-Wall -Wextra`, the remaining warnings are about the following point:
- Ignoring "'binder1st' is deprecated [-Wdeprecated-declarations]" in src/DGtal/geometry/surfaces/estimation/LocalEstimatorFromSurfelFunctorAdapter.h:145 because this function is deprecated since C++11.

#### General
I have modified the UNUSED macro in Common.h since it hides unused parameter only in Debug mode. The UNUSED_PARAM macro is strange because it lefts the documentation for unused parameter (see eg http://dgtal.org/doc/nightly/classDGtal_1_1SeparableMetricAdapter.html#a932f2272f78c976b483ffa4dc08c089f )


src/DGtal/geometry/curves/FrechetShortcut.h
```diff
@@ -584,9 +584,9 @@ namespace DGtal
 	 @param p and @param q two points
 	 @return an int
       */
-    static int computeQuadrant(Point p, Point q)
+    static int computeOctant(Point p, Point q)
```
dcoeurjo: ok I see what you meant.

src/DGtal/dec/DiscreteExteriorCalculusFactory.ih
```diff
@@ -144,9 +144,9 @@ DGtal::DiscreteExteriorCalculusFactory<TLinearAlgebraBackend, TInteger>::createF
     typedef typename Calculus::Cell Cell;
     typedef typename Calculus::SCell SCell;
     typedef typename Calculus::Scalar Scalar;
-    typedef typename Calculus::Point Point;
-    typedef typename Calculus::Property Property;
-    typedef typename Calculus::Properties Properties;
+    //typedef typename Calculus::Point Point;
+    //typedef typename Calculus::Property Property;
+    //typedef typename Calculus::Properties Properties;
```
dcoeurjo: these types should be removed if not used.

dcoeurjo:
I agree for UNUSED_PARAM, this is weird, it should remove the remaining line and not change the ```@param``` to ```param``` to prevent any warning with ```-Wdocumentation```

rolanddenis:
I've tried to change the macro but the problem is that doxygen doesn't interpret it. I think those macros was useful for variables that was only used in an ```ASSERT```, therefore generating warnings in release mode.
If we use the ```boost::ignore_unused_variable_warning``` macro, we can delete the ```UNUSED_*``` macros ... what do you think ?

src/DGtal/dec/DiscreteExteriorCalculusFactory.ih
rolanddenis: Ok, should I modify my pull request (since you have already merged it) ?

dcoeurjo:
Yeah UNSUED marcros are not great.. Should we promote following syntax instead ?:

     /**
      * blablabla
      * @param [in] exp the expression to parse (unused parameter)
      */
      void doSomething(const Expression &  exp)
      {
          boost:: ignore_unused_variable_warning(exp);

dcoeurjo:
would it be ok for both the compiler and doxygen ?

rolanddenis:
I was thinking about using this trick only if necessary but ok, it is easier if we used it everywhere.

However, we must be careful that ```boost/concept_check.hpp``` (or ```DGtal/base/Common.h```) is included when ```boost::ignore_unused_variable``` is used.

dcoeurjo:
all good. Merging

PR #1039, opened 2015-09-08
Linearizer

rolanddenis:
It provides linearization of point inside a given ```HyperRectDomain``` (with storage order specifier) and the reverse process (index to point). It replaces the linearizer used by ```ImageContainerBySTLVector``` and provides additional features:
- row-major and col-major storage order compatible (maybe not very useful ...);
- no loops, whatever the dimension ("meta-programming");
- possibility to be specialized for other domain types (if needed ?).

### DIFF
--- a/src/DGtal/kernel/domains/Linearizer.h
+++ b/src/DGtal/kernel/domains/Linearizer.h
@@ -1,0 +1,193 @@
+/**
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Lesser General Public License as
+ *  published by the Free Software Foundation, either version 3 of the
+ *  License, or  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ **/
+
+#pragma once
+
+/**
+ * @file Linearizer.h
+ * @author Roland Denis (\c roland.denis@univ-smb.fr )
+ * LAboratory of MAthematics - LAMA (CNRS, UMR 5127), University of Savoie, France
+ *
+ * @date 2015/06/18
+ *
+ * This file is part of the DGtal library.
+ */
+
+#if defined(Linearizer_RECURSES)
+#error Recursive header files inclusion detected in Linearizer.h
+#else // defined(Linearizer_RECURSES)
+/** Prevents recursive inclusion of headers. */
+#define Linearizer_RECURSES
+
+#if !defined Linearizer_h
+/** Prevents repeated inclusion of headers. */
+#define Linearizer_h
+
+//////////////////////////////////////////////////////////////////////////////
+// Inclusions
+#include <DGtal/kernel/domains/HyperRectDomain.h> // Only for specialization purpose.
+//////////////////////////////////////////////////////////////////////////////
+
+namespace DGtal
+{
+
+  /////////////////////////////////////////////////////////////////////////////
+
+  /**
+   * @brief Tag (empty structure) specifying a row-major storage order.
+   *
+   * @see Linearizer
+   */
+  struct RowMajorStorage {};
+
+  /**
+   * @brief Tag (empty structure) specifying a col-major storage order.
+   *
+   * @see Linearizer
+   */
+  struct ColMajorStorage {};
+
+  /////////////////////////////////////////////////////////////////////////////
+  /**
+   * @brief Aim: Linearization and de-linearization interface for domains.
+   *
+   * Description of class 'Linearizer' <p>
+   * This class is empty but there is a specialization for HyperRectDomain.
+   *
+   * @tparam  TDomain       Type of the domain.
+   * @tparam  TStorageOrder Storage Order (RowMajorStorage of ColMajorStorage).
+   *
+   */
+  template <
+      typename TDomain,
+      typename TStorageOrder = ColMajorStorage
+    >
+  struct Linearizer;
+
+  /**
+   * @brief Aim: Linearization and de-linearization interface for HyperRectDomain.
+   *
+   * This is a static class that provides point linearization (point to index) and de-linearization (index to point) for storages working on HyperRectDomain.
+   *
+   * The storage order can be specified by template (default is colum-major ordered).
+   *
+   * Example:
+   * @code
+   * typedef SpaceND<2>             Space;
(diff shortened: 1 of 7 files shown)


dcoeurjo:
unit test of this class ;)  ? For that, you can play with the brand new ```catch```framework to write quick tests.
(it may be difficult since you've removed the old linearizer but I'd love to see performance comparisons..)

src/DGtal/kernel/domains/Linearizer.h
```diff
@@ -0,0 +1,193 @@
...
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#endif // !defined Linearizer_h
+
+#undef Linearizer_RECURSES
+#endif // else defined(Linearizer_RECURSES)
+
+/* GNU coding style */
+/* vim: set ts=2 sw=2 expandtab cindent cinoptions=>4,n-2,{2,^-2,:2,=2,g0,h2,p5,t0,+2,(0,u0,w1,m1 : */
```
dcoeurjo: cleanup these editor traces?

src/DGtal/kernel/domains/Linearizer.h
```diff
@@ -0,0 +1,193 @@
...
+   *
+   * Example:
+   * @code
+   * typedef SpaceND<2>             Space;
+   * tydedef HyperRectDomain<Space> Domain;
+   * typedef typename Space::Point  Point;
+   *
+   * const Domain domain( Point(0, 1), Point(4, 3) );
+   *
+   * size_t id = Linearizer<Domain>::getIndex( Point(2, 2), domain ); // returns 7.
```
dcoeurjo: shouldn't it be 
       Linearizer<Domain>::Size id = .....
?

src/DGtal/kernel/domains/Linearizer.ih
```diff
@@ -0,0 +1,224 @@
...
+
+namespace DGtal
+{
+
+  /// Tools
+  namespace
+  {
+
+    /** Helper that calculates the dimension index given the current linearization step.
+     *
```
dcoeurjo: very nice template expression programing  ;)

src/DGtal/kernel/domains/Linearizer.ih
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+  /// De-linearization of an index, given a domain.
+  template <typename TSpace, typename TStorageOrder>
+  typename Linearizer< HyperRectDomain<TSpace>, TStorageOrder >::Point
+  Linearizer< HyperRectDomain<TSpace>, TStorageOrder >::
+      getPoint( Size anIndex, Domain const& aDomain )
+    {
+      Point point;
+      linearizer_impl<Size, TStorageOrder, Domain::dimension>::apply(point, aDomain.upperBound()-aDomain.lowerBound()+Point::diagonal(1), anIndex);
```
dcoeurjo: don't you think "caching" the ```aDomain.upperBound()-aDomain.lowerBound()+Point::diagonal(1)``` would considerably improve the computations? 
it would require data member...

src/DGtal/kernel/domains/Linearizer.ih
```diff
@@ -0,0 +1,224 @@
...
+      return point + aDomain.lowerBound();
+    }
+
+} // namespace DGtal
+
+/* GNU coding style */
+/* vim: set ts=2 sw=2 expandtab cindent cinoptions=>4,n-2,{2,^-2,:2,=2,g0,h2,p5,t0,+2,(0,u0,w1,m1 : */
```
dcoeurjo: cleanup?

dcoeurjo:
(changelog entry ?)

src/DGtal/kernel/domains/Linearizer.h
rolanddenis: Oooops, I will clean this !

src/DGtal/kernel/domains/Linearizer.h
rolanddenis: Right, with probably a ```typename``` before.

src/DGtal/kernel/domains/Linearizer.ih
rolanddenis: Thansk ! It was funny to write it ;p

src/DGtal/kernel/domains/Linearizer.ih
rolanddenis: It will probably improve the computations if used many times but it was meant to stay a static class (to be as simple as possible). I propose to add in the documentation that the methods that use the extent parameter are more efficient.

... otherwise, we could add an additional non-static class meant for "heavy use" and constructible from the domain or the extent&lowerBound. What do you think ?

src/DGtal/kernel/domains/Linearizer.ih
rolanddenis: Yep!

rolanddenis:
Concerning the unit test, I will look at it ! 
About the speed improvement, I don't think that it will be a big difference because:
- I think the compiler unrolls the loops of the previous version (since he knows the number of iterations);
- this version calculates, before linearisation, the difference between the point and the lower bound of the domain, while the previous version calculates it during the linearisation. (is there a speed difference ?)

rolanddenis:
Done. I think that it tests enough different cases, don't you ? :D 
An benchmark is accessible through hidden tag [.bench], eg:
```./kernel/testLinearizer "[.bench]" -d yes```
Since each test case (in benchmark) includes a warmup step, always look at odd lines for timing (corresponding to a section).

Tags are available for each space dimension (up to 5) and each storage order, eg:
```./kernel/testLinearizer "[.bench][dim3][ColMajorStorage]" -d yes```

File: tests/kernel/testLinearizer.cpp [hunk 6]
```diff
@@ -0,0 +1,349 @@
...
+      TPoint result;
+      for ( std::size_t i = 0 ; i < TPoint::dimension ; ++i )
+        result[i] = aPoint[ TPoint::dimension - i - 1 ];
+
+      return result;
+    }
+};
+
+#define TEST_LINEARIZER( N , ORDER ) \
+TEST_CASE( "Testing Linearizer in dimension " #N " with " #ORDER, "[test][dim" #N "][" #ORDER "]" )\
```
dcoeurjo: :smile:  nice macro based test case generator ;)
unfortunately, there is no such thing in catch yet https://github.com/philsquared/Catch/issues/358

or https://github.com/philsquared/Catch/issues/46

dcoeurjo:
thanks a lot !
merging

tests/kernel/testLinearizer.cpp
rolanddenis: And there is also a pull request about templated test case: philsquared/Catch#382
It is maybe possible to do cartesian product with macro ? ;)

dcoeurjo: Yeah I saw this PR too (but stalled since many months)... and for me, macro programming is like juggling with activated grenades.. :smile:

PR #1040, opened 2015-09-09
Removing coveralls for code coverage

dcoeurjo:
due to coveralls.io issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: cpp
 
 sudo: false
 
-
 compiler:
   - gcc
   - clang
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -11,8 +10,6 @@ os:
   - linux
   - osx
 
-osx_image: xcode61
```
kerautret: it is no more used? or can be updated with 6.4 or 7.0 ?

dcoeurjo: I think we can use different version of Xcode. By removing this, I just wanted to use the default setting (no need in dgtal).

kerautret: ok.
By the way it is very fine the potential of DGtal in xcode (in the first use Xcode takes a little time to parse headers but after it is nice)...

PR #1045, opened 2015-10-02
Fix for: CConstSinglePassRange.h:100:21: error: variable ‘it’ set but…

copyme:
… not used [-Werror=unused-but-set-variable]

One Travis test for my branch LambdaMST failed because of this error. I could not reproduce it here but I believe that this should fix the problem.


dcoeurjo:
Agree.. Many thanks

Merging

PR #1055, opened 2015-10-17
Dev catch

dcoeurjo:
* Upgrade to catch develop version 
* fixing @kerautret catch test ( "&&" in REQUIRE does not compile anymore, please use separate REQUIRE)
* The dedicated Catch option is set if the  build is without CPP11.

poke @rolanddenis


kerautret:
ok for the && ;) merging thanks

dcoeurjo:
Thanks. 
@kerautret, please have a look to point vector unit test, catch also provides an automatic ```approx``` to avoid your 0.000001 constants

kerautret:
yes I check it that's great the approx() ! I change it in AlphaThickPR

kerautret:
thanks!

PR #1060, opened 2015-10-22
Enabling and testing circulators in SimpleRandomAccess(Const)RangeFromPoint

rolanddenis:
Circulators seem to have been disabled in `SimpleRandomAccessRangeFromPoint` but I don't know why ... Since they are already available in `SimpleRandomAccessConstRangeFromPoint`, this pull request adds it for the mutable ranges.

A test file has been also added and the discovered bugs in reverse iterators have been fixed.


rolanddenis:
There is an (small) issue with gcc 4.6.3: it always suggests adding parentheses around comparison in macro, even if it is not needed. It seems to be the only one to complain about that ...

The problem is that Catch cannot expand failed assertions if those parentheses are added.
For example, with the snippet
```
int a = 1, b = 2;
REQUIRE( a == b );
```
Catch output the following message:
```
test.cpp:123: FAILED:
  REQUIRE( a == b );
with expansion:
  1 == 2
```
but with added parentheses:
```
int a = 1, b = 2;
REQUIRE(( a == b ));
```
it outputs:
```
test.cpp:123: FAILED:
  REQUIRE( ( a == b ) );
with expansion:
  false
```

We should maybe use a more recent compiler or disabling the `WARNING_AS_ERROR` option for this version ? (I admit this it is not a critical issue ...)

dcoeurjo:
Oh I see... I've  also added extra parenthesis un my tests for travis but I didn't know that it breaks the expansion output (which is one of the main argument for catch)... 
Would it be possible to disable only this parenthesis warning only for this version of GCC ?

dcoeurjo:
-Werror=parentheses   if the flag that should be "unset"

rolanddenis:
As far as I understand Catch, it uses the operator precedence to build expression template on the tested expression. For example, if the given expression in REQUIRE macro is `2 >= 3`, Catch replaces it by
```ResultBuilder <= 2 >= 3```
As `<=` and `>=` have same priority, `C++` evaluates firstly `ResultBuilder <= 2` that allows it to capture 2, then he evaluates `ResultedExpression >= 3` that allows it to capture the comparison operator and 3.

The following snippet reproduces the basic behaviour of Catch:
```
#include <iostream>

using namespace std;

struct Dummy
{
    template <typename T>
    Dummy const& operator<= (T const& rhs) const
    {
        cout << " <= " << rhs;
        return *this;
    }

    Dummy const& operator<= ( bool rhs ) const
    {
        cout << " <= " << ( rhs ? "true" : "false" );
        return *this;
    }
    
    template <typename T>
    Dummy const& operator== (T const& rhs) const
    {
        cout << " == " << rhs;
        return *this;
    }

    // And so on with other comparison operators.
};

int main()
{
    Dummy() <= 1*2 == 5-1;
    cout << " ?" << endl;

    Dummy() <= (1*2 == 5-1);
    cout << " ?" << endl;

    return 0;
}
```
that output
```
<= 2 == 4 ?
<= false ?
```
(a additional indirection is needed in order to avoid the first `<=`)

In that case, adding parentheses around the expression breaks the operator precedence and the compiler evaluates `2 == 4` before `Dummy() <= ...`.

In my opinion, it is normal that the compiler warns us when he reads an expression like `Dummy() <= 1*2 == 5-1` because there can be a misunderstanding. `g++-4.9.2` and `g++-4.6.4` emit a warning on the previous snippet only with `-Wall` flag (that contains `-Wparentheses` ) but `clang 3.6.0` never complains.

Thus, what I don't understand is how Catch avoids those warnings with `g++` other than `4.6`, and where the `-Wparentheses` flag is set ...

**Comment by rolanddenis** [comment]
Putting this two lines (see the last commit):
```
_Pragma("GCC diagnostic push")
_Pragma("GCC diagnostic ignored \"-Wparentheses\"")
```
before, and this line:
```
_Pragma("GCC diagnostic pop")
```
after the line that generates the warning solves our problem.
`_Pragma` is used instead of `#pragma` because we are in a macro definition (see e.g. http://stackoverflow.com/questions/3030099/pragma-in-define-macro).

What I don't understand is that it is the line
```} while( Catch::isTrue( false && (expr) ) )```
that seems to generate the warning (even if `expr` has parentheses around) and not the line (in the loop):
```( __catchResult <= expr ).endExpression(); \```
as I expected :expressionless:

I have tested several gcc versions :
- 4.6.4 and 4.7.4 trigger the warning,
- 4.8.4 and 4.9.2 are ok.

Should I continue (maybe in an other PR) ?

rolanddenis:
Funny: if I replace the `while` condition by
```} while( Catch::isTrue( false ) )```
and remove the pragmas, there is still a warning :smile: 
So it must be the ```( __catchResult <= expr ).endExpression(); \``` line that triggers the warning but the pragmas have a strange behavior ...

rolanddenis:
Ok, there is multiple bug reports about `_Pragma` :confused:
https://www.mail-archive.com/gcc-bugs@gcc.gnu.org/msg457701.html
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66099
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53469
...

rolanddenis:
I have move the discussion to #1067 and revert the previous changes.

dcoeurjo:
Beside Catch issues (addressed in #1067), @troussil , could you please have a look to iterator adapters provided in this PR ?
(thanks @rolanddenis for the PR)

troussil:
when compiling with gcc 4.8 I have a bunch of warnings because SpaceND::Size is unsigned (and thus domain.size() is unsigned):

eg. 
DGtal/tests/base/testSimpleRandomAccessRangeFromPoint.cpp:69:60: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
       REQUIRE(( range.end() - range.begin() == domain.size() ));

rolanddenis:
Thanks ! I forgot to cast to unsigned before comparing with domain size.
Strangely, none of my compilers complained about that ...

troussil:
Sorry, but it remains one comparison between signed and unsigned integer line 73...

rolanddenis:
Done :)

troussil:
thanks

PR #1066, opened 2015-11-01
Minor Viewer3D improvements

dcoeurjo:
- Default background is lighter
- The light position has been changed so that it is not axis aligned anymore 
- default window size is 800x600


dcoeurjo:
BTW, I don't understand why ```viewer3D-5-custom``` points have black edges..

src/DGtal/io/viewers/Viewer3D.ih
```diff
@@ -820,7 +820,7 @@ DGtal::Viewer3D<Space, KSpace>::init()
   myGLScaleFactorY=1.0;
   myGLScaleFactorZ=1.0;
   myGLLineMinWidth=1.5;
-
+  setBackgroundColor ( QColor ( 217, 228, 255 ) );
```
kerautret: For the color ok but the dark one is no more present. 
We can perhaps put the new one by default (instead white) and let the dark one associated to Key_B ?

kerautret:
else looks fine thanks

src/DGtal/io/viewers/Viewer3D.ih
dcoeurjo: The white background is still needed for exports and snapshots. 
The idea of this PR is to replace the dark background (which is not great when displaying isothetic quads with classical lightning) by a "lighter" one.

dcoeurjo:
Does the resize to 800x600 works on your machine?

src/DGtal/io/viewers/Viewer3D.ih
kerautret: ok I see and we could later add a third 'dark' mode if needed.

kerautret:
yes it works, not you ? (at least for the examples)

dcoeurjo:
not always working.. strange...
Ready to merge?

jlevallois:
It's ok for me

dcoeurjo:
ok merging. thx

PR #1069, opened 2015-11-03
Fix parentheses warning with Catch (again)

rolanddenis:
Second chance for the PR #1067.

The problem is now with expression like
```
REQUIRE( 1+2 == 3 );
```
that emits warning on adding parentheses around ```1+2```. 

It still allows Catch to expand each side of the comparison in case of assertion failure but it forces us to add extra-parentheses when left term contains `+` or `-`. It's more or less the same as when Catch was using `->` operator (but compilation was failing without the extra-parentheses). 

We can also keep this PR as is and wait for an official fix.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -24,8 +24,8 @@
  - Fixing boost include path issue when building tools using DGtal and
    its cmake DGtalConfig.cmake. (David Coeurjolly,
    [#1059](https://github.com/DGtal-team/DGtal/pull/1059))
- 
-
+ - Fixing parenthese warnings in Catch. Waiting for an official fix.
+   (Roland Denis, [#1069](https://github.com/DGtal-team/DGtal/pull/1069))
 
 - *Base Package*
  - Fix wrong initialization of reverse iterators in SimpleRandomAccess(Const)RangeFromPoint.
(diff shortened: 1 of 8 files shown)


dcoeurjo:
Thanks a lot @rolanddenis, let's consider this solution. 
Can you please add an entry to the changelog?

rolanddenis:
Thus we stay on the exact same PR than previously ?

dcoeurjo:
yep (1+2)==5 is way better than ((1+2)==5)
does it generate warnings on your compilers?

dcoeurjo:
(but prev PR was reverted)

rolanddenis:
no, adding parentheses around `+` and `-` avoids the warnings. But if we accept to add those parentheses, we could also go back to `->*` that is surely less prone to generate warnings (less precedence) and that clearly failed to compile (with a message, see https://github.com/DGtal-team/DGtal/blob/master/tests/catch.hpp#L1867 ) instead of having a warning on some compiler.

dcoeurjo:
Up to you.. you're the specialist;)

rolanddenis:
Hum ok, we stay on this PR and we will see if someone complains :)

rolanddenis:
Done. I have also remove extra parentheses from testSimpleRandomAccessRangeFromPoint.

dcoeurjo:
:cool: :ok_hand:

PR #1070, opened 2015-11-03
Move light position

kerautret:
add a new possibility to move light source position (an x-axis (resp. y-axis) mouse move changes the azimuth (resp. inclination angle) of the light source).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,7 +13,7 @@
  - Minor improvements of default settings in Viewer3D. (David
    Coeurjolly, [#1066](https://github.com/DGtal-team/DGtal/pull/1066))
 
- - New possibility to move the light source according to the mouse move in
+ - New possibility to move the light source using the mouse move in
    Viewer3D.
    (Bertrand Kerautret [#1070](https://github.com/DGtal-team/DGtal/pull/1070))
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -14,7 +14,7 @@
    Coeurjolly, [#1066](https://github.com/DGtal-team/DGtal/pull/1066))
 
  - New possibility to move the light source using the mouse move in
-   Viewer3D.
+   Viewer3D (with the key SHIFT+CTRL (SHIFT+CMD on mac)).
    (Bertrand Kerautret [#1070](https://github.com/DGtal-team/DGtal/pull/1070))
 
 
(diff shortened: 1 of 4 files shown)


kerautret:
The light source direction is also displayed during the light source move.

ChangeLog.md
```diff
@@ -13,6 +13,11 @@
  - Minor improvements of default settings in Viewer3D. (David
    Coeurjolly, [#1066](https://github.com/DGtal-team/DGtal/pull/1066))
 
+ - New possibility to move the light source according to the mouse move in
```
dcoeurjo: ... source using the mouse (with SHIFT key modifier) in Viewer3D

src/DGtal/io/viewers/Viewer3D.h
```diff
@@ -1267,7 +1293,11 @@ namespace DGtal
     float myGLScaleFactorX;
     float myGLScaleFactorY;
     float myGLScaleFactorZ;
-
+    // Used to apply interactive light rotation
+    float myLigthRotationStep;
+    int myRefMouseXPos;
+    int myRefMouseYPos;
```
dcoeurjo: doxygen doc of the members ?

dcoeurjo:
Nice feature ;) i'll test it

ChangeLog.md
kerautret: better thanks ;)

src/DGtal/io/viewers/Viewer3D.h
kerautret: ok

kerautret:
done ;)

dcoeurjo:
x-y seems to be ok but I don't understand the "z"
BTW add the key binding (shift+CMD on mac..) in the changelog
and add something to say that the light always looks at (0,0,0)

kerautret:
the position z of the light should change when you move the mouse according an y move (upper or lower). 
ok for the key

dcoeurjo:
euh.. I don"t get it.. Are x (resp. y) moves increasing the phi angle (resp. theta) in a Euler representation ?

kerautret:
yes exactly

kerautret:
I fix a pb of direction update from the draw(), now it should be finer. The light source direction is always defined in the main system coords, and don't move with the camera. I think that this choice is better to simplify the user interaction with the definition of move.
Hope that it is also fine for you?

dcoeurjo:
perfect!! thanks a lot

dcoeurjo:
Would it be possible to add a line in the documentation ?

kerautret:
it in the help function but you mean in the main section ok I add it

dcoeurjo:
I meant in the module documentation

kerautret:
done

dcoeurjo:
great
merging

PR #1072, opened 2015-11-05
Catch link

dcoeurjo:
Adding a symbolic link from ```tests/catch.hpp``` to ```examples/doc-examples/catch.hpp```


dcoeurjo:
Doing a self merge, nothing special.

PR #1078, opened 2015-11-21
Fix random

dcoeurjo:
Just a workaround in KanungoNoise to fix the PSNRG.

This fix is a bit complex since the boost API has changed between 1.46 (boost version in travis) and  1.47 (and later).
Furthermore, if c++11 is enabled, the std::random framework is used.


dcoeurjo:
ping @JacquesOlivierLachaud  & @copyme

cmake/CheckDGtalDependencies.cmake
```diff
@@ -14,9 +14,17 @@ set(Boost_USE_STATIC_RUNTIME OFF)
 set(Boost_FOUND FALSE)
 FIND_PACKAGE(Boost 1.46.0 REQUIRED)
 if ( Boost_FOUND )
-  message(STATUS "Boost found.")
   include_directories( ${Boost_INCLUDE_DIRS} )
   SET(DGtalLibInc ${DGtalLibInc} ${Boost_INCLUDE_DIRS})
+
+  ## Checking boost/random ( <1.47)
+  STRING (COMPARE LESS "${Boost_MINOR_VERSION}" 47 BOOST_RANDOM_OLD)
```
rolanddenis: What will happen when Boost reaches version 2 (one day) ? Isn't it safer to use `BOOST_VERSION` available in `FindBoost`:
```STRING (COMPARE LESS "${Boost_VERSION}" 104700 BOOST_RANDOM_OLD)```

dcoeurjo: I tried something based on Boost_VERSION but it didn't work (not sure how cmake really compares strings with numerals). I'll try your line, thx

dcoeurjo:
works like a charm ;)

JacquesOlivierLachaud:
Seems fine to me. However, I would not mind a DGtal::functions::uniform_random(), which would avoid all these preprocessor macros each time we need a simple uniform random generator.

dcoeurjo:
yes I agree, the preprocessing comments are insane.. However, if we decide to switch to c++11 for the lib (see working branch https://github.com/dcoeurjo/DGtal/tree/cpp11), it would be nt needed anymore..

dcoeurjo:
(selfmerge)

PR #1079, opened 2015-11-24
Cubical complex

JacquesOlivierLachaud:
This PR includes:
- new traits class for Container to detect their category
- a module for performing set operations on arbitrary containers (base)
- a module for representing cubical complex and operations associated (incidence, iteration, link, closure, star, set operations, collapse), as well as display on stream.
- some examples


JacquesOlivierLachaud:
Almost ready. Need a few update in the doc.

bmart36:
I think your code couldn't be compiled because you used std::move which was introduced in cpp11

src/DGtal/topology/CubicalComplexFunctions.h
```diff
@@ -0,0 +1,413 @@
...
+       * (but the constant is also linear in the dimension of \a K).
+       */
+      template <typename TCubicalComplex,
+                typename CellOutputIterator>
+      void 
+      filterCellsWithinBounds( const TCubicalComplex& K, 
+                               const typename TCubicalComplex::Point& kLow,  
+                               const typename TCubicalComplex::Point& kUp,
+                               CellOutputIterator itBdry, 
+                               CellOutputIterator itInner );
```
rolanddenis: I suggest to use two different template types for the two output iterators. It allows the use of different containers, or to use an output iterator that do nothing, or to apply a function on each cells (see e.g. `boost::function_output_iterator`).

dcoeurjo:
Let me start with general remarks.

ContainerTraits: 
* please remove c++11 #ifdef and use std containers as well (e.g. std::unordered_set), the c++11-only PR has been merged.
* do you suggest to rewrite Container Concept Checks we have with Traits based static assert  (e.g. ``CSTLAssociativeContainer.h``` or boost concept checks) ?

tests/base/testSetFunctions.cpp
```diff
@@ -0,0 +1,337 @@
...
+using namespace DGtal::functions;
+using namespace DGtal::functions::setops;
+
+////////////////////////////// unit tests /////////////////////////////////
+TEMPLATE_TEST_CASE_4( "SetFunctions module unit tests", "[set_functions]",
+                      Container,
+                      std::vector<int>,
+                      std::list<int>, 
+                      std::set<int>, 
+                      boost::unordered_set<int> )
```
dcoeurjo: std::unordered_set (please check elsewhere)

src/DGtal/base/SetFunctions.h
```diff
@@ -0,0 +1,1469 @@
...
+     *
+     *
+     * Specialized implementations are marked with (S) in the list below.
+     *
+     * |--------------------|-------------|----------|----------|
+     * | Container          | associative | ordered  |  pair    |
+     * |--------------------|-------------|----------|----------|
+     * | vector             |   false     |  false   |  false   |
+     * | list               |   false     |  false   |  false   |
+     * | (to avoid)         |   false     |  false   |  true    |
```
dcoeurjo: i don't get the ```(to avoid)```

src/DGtal/base/SetFunctions.h
```diff
@@ -0,0 +1,1469 @@
...
+     * data is not taken into account, which means that it can be lost
+     * in some (modifier) operations.
+     */
+    template <typename Container, bool associative, bool ordered>
+    struct SetFunctionsImpl
+    {
+      /** 
+       * Equality test. This version does not use the
+       * fact that the container is ordered.
+       * @param[in] S1 an input set.
```
dcoeurjo: maybe you should state that container values are be sorted (with O(nlogn) cost for some containers) before comparing them.

src/DGtal/base/SetFunctions.h
```diff
@@ -0,0 +1,1469 @@
...
+        std::sort( V1.begin(), V1.end(), CompAdapter::less( S1 ) );
+        std::sort( V2.begin(), V2.end(), CompAdapter::less( S1 ) );
+        return std::equal( V1.begin(), V1.end(), V2.begin(),
+                           CompAdapter::equal_to( S1 ) );
+      }
+
+      /** 
+       * Inclusion test. This version does not use the
+       * fact that the container is ordered.
+       * @param[in] S1 an input set.
```
dcoeurjo: same here about the inner std::sort s

ChangeLog.md
```diff
@@ -2,6 +2,21 @@
 # DGtal 0.9.1
 
 ## New Features / Critical Changes
+- *Base Package*
+ - Traits class for containers in order to probe their category at compile time.
+   (Jacques-Olivier Lachaud, [#1079](https://github.com/DGtal-team/DGtal/pull/1079))
+ - Generic set operations for arbitrary containers. You may use overloaded operators
+   like &, |, -, ^ on arbitrary containers (list, vector, unordered_set, map, etc).
+   (Jacques-Olivier Lachaud, [#1079](https://github.com/DGtal-team/DGtal/pull/1079))
+
```
dcoeurjo: Can DigitalSet containers could also be considered in these specializations ? It seems to be pretty trivial for the ```DigitalSetByAssociativeContainer``` class for instance.

dcoeurjo: it would be great to be able to use set operators on digital point sets (static domains could be problematic btw..)

dcoeurjo:
could you please merge the master to this branch ? (C++11 stuff is conflicting).

src/DGtal/base/doc/moduleSetFunctions.dox
```diff
@@ -0,0 +1,189 @@
...
+It is enough to include module SetFunctions.h and then write `using
+namespace DGtal::functions::setops` to use binary operators directly
+on your containers. The following snippet shows an example:
+
+\code
+#include "DGtal/base/SetFunctions.h"
+...
+using namespace DGtal::functions::setops;
+
+typedef std::list<int> Container; // could be boost::unordered_set<int>, etc
```
dcoeurjo: boost->std

src/DGtal/base/doc/moduleSetFunctions.dox
```diff
@@ -0,0 +1,189 @@
...
+@section dgtal_setfunctions_sec2  Performing set operations
+
+It is enough to include module SetFunctions.h and then write `using
+namespace DGtal::functions::setops` to use binary operators directly
+on your containers. The following snippet shows an example:
+
+\code
+#include "DGtal/base/SetFunctions.h"
+...
+using namespace DGtal::functions::setops;
```
dcoeurjo: At some point, we discussed about "sub-namespaces" in ```DGtal::functions, DGtal::concepts, .... ``` and the conclusion was to avoid them.
Do you suggest to allow them?

dcoeurjo: I understand why you need a specific namespace for operator overloading but that's a bit strange as a subnamespace of ```functions```

src/DGtal/base/SetFunctions.h
dcoeurjo: Ok this is checked in the specialized version. But it still should be indicated in the "non"-specialized documentation (that doxygen will consider)

src/DGtal/base/doc/moduleSetFunctions.dox
```diff
@@ -0,0 +1,189 @@
...
+| set<int>     union     |      2.0 |     25.8 |    308.7 |   3402.6 |
+| (conversion) union     |      2.1 |     27.0 |    322.3 |   3559.7 |
+| set<int>     inter     |      1.5 |     17.3 |    202.0 |   2269.4 |
+| (conversion) inter     |      1.7 |     19.3 |    241.8 |   2515.9 |
+| set<int>     diff      |      1.3 |     14.5 |    171.0 |   1821.7 |
+| (conversion) diff      |      1.4 |     16.5 |    203.3 |   2135.7 |
+| set<int>     sym_diff  |      1.7 |     18.6 |    213.8 |   2403.7 |
+| (conversion) sym_diff  |      1.6 |     20.0 |    242.6 |   2660.7 |
+|                        |          |          |          |          |
+| un_set<int>  union     |      1.0 |     10.5 |    176.2 |   2464.1 |
```
dcoeurjo: unordered_set ?

dcoeurjo: please expand.

src/DGtal/io/Display2DFactory.ih
```diff
@@ -1453,6 +1453,39 @@ void DGtal::Display2DFactory::draw( DGtal::Board2D & board,
...
+  typedef typename CC::Cell                 Cell;
+  typedef typename CC::CellMapConstIterator CellMapConstIterator;
+  typedef typename CC::ConstIterator        ConstIterator;
+  typedef typename KSpace::Integer          Integer;
+  for ( ConstIterator it = cc.begin(), itE = cc.end(); it != itE; ++it )
+    { // We are not using operator << for cells in order to use the
+      // style specified for the cubical complex.
+      Point kx = cc.space().uKCoords( *it );
+      float x = (float) ( NumberTraits<Integer>::castToInt64_t( kx[0] ) >> 1 );
+      float y = (float) ( NumberTraits<Integer>::castToInt64_t( kx[1] ) >> 1 );
```
dcoeurjo: why not using the CanonicalCellEmbedder ?
It's better to have the embedding conventions at the same place.

src/DGtal/io/Display2DFactory.ih
```diff
@@ -1453,6 +1453,39 @@ void DGtal::Display2DFactory::draw( DGtal::Board2D & board,
...
+      float retract = 0.05f;
+      bool xodd = ( kx[ 0 ] & 1 );
+      bool yodd = ( kx[ 1 ] & 1 );
+      float dx = xodd ? 1.0f - 2.0f*retract : 2.0f*retract;
+      float dy = yodd ? 1.0f - 2.0f*retract : 2.0f*retract;
+      board.drawRectangle( !xodd ? x - retract - 0.5 : x + retract - 0.5,
+                           !yodd ? y + retract - 0.5 : y - retract + 0.5,
+                           dx, dy );
+    }
```
dcoeurjo: can't you reuse KSCell/KCell drawing mechanisms?

dcoeurjo:
(I will continue later)

tests/base/testSetFunctions.cpp
JacquesOlivierLachaud: I will do it everywhere.

src/DGtal/base/SetFunctions.h
JacquesOlivierLachaud: In case somebody use a weird container for set: non associative, non sorted, and with a pair key/value. Like an unsorted list of pairs. Not a reasonable choice for a set structure.

src/DGtal/base/SetFunctions.h
JacquesOlivierLachaud: In fact, it does sort data before comparison in the non-specialized version, so the user does not need to sort data before (and sometimes, he cannot, e.g. unordered_set). Have I misunderstood something ?

ChangeLog.md
JacquesOlivierLachaud: I must check but I don't think it is so easy to do it. I think we will have constructor problems (no default constructor since we need a domain to create a set). Perhaps it is possible with the DigitalSetByAssociativeContainer, by using the copy constructor (sorted specialization does not use default constructor).

src/DGtal/base/doc/moduleSetFunctions.dox
JacquesOlivierLachaud: I did not want to pollute the global or local namespace. For instance, such operator overloading interferes with catch operator overloading.

src/DGtal/base/doc/moduleSetFunctions.dox
JacquesOlivierLachaud: done.

src/DGtal/io/Display2DFactory.ih
JacquesOlivierLachaud: I do not use the SCell/Cell mechanism because I wish that the user can specify a style for CubicalComplex. I agree that the display part is common. I can try to factor it.

src/DGtal/io/Display2DFactory.ih
JacquesOlivierLachaud: I agree, but it was not also used in SCell/Cell I think.

dcoeurjo: Yep but it should.

cmake/CheckDGtalOptionalDependencies.cmake
```diff
@@ -124,6 +124,61 @@ ENDIF(WITH_BENCHMARK)
 message(STATUS "")
 message(STATUS "Checking the dependencies: ")
 
+# # -----------------------------------------------------------------------------
+# # Check CPP11
```
dcoeurjo: Please cleanup

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1305 @@
...
+#include "DGtal/topology/CCellularGridSpaceND.h"
+//////////////////////////////////////////////////////////////////////////////
+
+namespace DGtal
+{
+
+  /**
+  * Any cell is stored within a cubical complex with an associated
+  * data, which must derive from this class. Its basic usage is to
+  * store flags associated to the cells, but it may store other
```
dcoeurjo: The first sentence is weird.
Remove the which ?

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1305 @@
...
+  };
+  
+  template < typename TKSpace, typename TCellContainer >
+  class CubicalComplex;
+  
+  namespace functions {
+    namespace ccops {
+      template < typename TKSpace, typename TCellContainer >
+      CubicalComplex< TKSpace, TCellContainer >&
+      operator|=( CubicalComplex< TKSpace, TCellContainer >&,
```
dcoeurjo: Since these operators are related to ANR explicit (but templated) class Cubicalcomplex, i don't see the interest o move the operators to a specific namespace. (The set functions cases is a bit different since we may also use == operator on std vectors without considering them as set containers)

dcoeurjo:
I have the feeling that comparison operator overloading in SetFunctions could be very dangerous: if someone use the setops namespace, all comparison operators on PointVectors for instance would be completely screwed since  PointVector ```==```relies on  ```==``` operator of std::array (same for ```<```). This would have a huge sideffects in many part of DGtal. 

What do you think?

I would suggest to keep set operators ```!,&,...``` but remove comparison ones (which use a specific "set functions" semantic).

dcoeurjo:
(I'm not sure what the compiler does when there are several operator overloading in different namespaces).

src/DGtal/base/doc/moduleSetFunctions.dox
dcoeurjo: Yeah and also in PointVector (see comment below)

JacquesOlivierLachaud: I have removed <=, ==, != and >= operators for sets and update the doc.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: I have moved all cubical complex functions into namespace functions.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: Fixed.

cmake/CheckDGtalOptionalDependencies.cmake
JacquesOlivierLachaud: done

src/DGtal/io/Display2DFactory.ih
JacquesOlivierLachaud: I prefer not to do it in this PR, but to do it properly in a separate PR, along with 3D view. I have the feeling that I may run into trouble (because of shifts problem). So perhaps we can raise an issue here.

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1305 @@
...
+  {
+    // ----------------------- associated types ------------------------------
+  public:
+    typedef CubicalComplex< TKSpace, TCellContainer > Self;
+    
+    BOOST_CONCEPT_ASSERT(( concepts::CCellularGridSpaceND< TKSpace > ));
+    BOOST_STATIC_ASSERT( IsPairAssociativeContainer< TCellContainer >::value );
+
+    // JOL: Not used, because unordered_set and unordered_map do not pass
+    // these concept checks.
```
dcoeurjo: Agree. Maybe you can use the concept  I have created ```CSTLAssociativeContainer```

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1305 @@
...
+
+    /// Possible cell types within a complex. 
+    enum CellType { 
+      Maximal /**< The cell has no proper coface */, 
+      Free    /**< The cell has 1 proper coface */,
+      Any     /**< The cell has strictly more than 2 proper cofaces.*/
+    };
+
+    /// Flag Used to indicate in a cell data that this cell has been (virtually) removed.
+    static const uint32_t REMOVED     = 0x10000000;
```
dcoeurjo: shouldn't you use ```BOOST_STATIC_CONSTANT```? (http://www.boost.org/doc/libs/1_31_0/more/int_const_guidelines.htm)

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1305 @@
...
+    typedef TCellContainer CellContainer;
+    typedef typename CellContainer::mapped_type Data;
+
+    BOOST_STATIC_ASSERT (( boost::is_base_of< CubicalCellData, Data >::value ));
+    BOOST_STATIC_ASSERT (( boost::is_same< typename TKSpace::Cell, typename CellContainer::key_type >::value ));
+
+
+    /// The dimension of the embedding space.
+    static const Dimension dimension = KSpace::dimension;
+    typedef typename KSpace::Integer     Integer;
```
dcoeurjo: doxygen doc of these types?

src/DGtal/topology/doc/moduleCubicalComplex.dox
```diff
@@ -0,0 +1,328 @@
...
+testCubicalComplex.cpp,
+
+@section dgtal_ccomplex_sec1  Introduction to cubical complexes
+
+We define a \b cubical \b complex \a C as a collection of cells living
+in some Khalimsky space. Two cells of \a C are \b incident if and only
+if they are incident in the Khalimsky space. 
+
+@note In opposition to the usual definition of simplical complex, we
+do not require that all faces of a cell of \a K belong also to \a K.
```
dcoeurjo: K -> C since you refer to C

src/DGtal/topology/doc/moduleCubicalComplex.dox
```diff
@@ -0,0 +1,328 @@
...
+
+To create a cubical complex, we need to specify in which Khalimsky
+space it lives and also, optionally, the type of container used for
+storing cells. By default it is \c std::map but \c boost::unordered_map
+or \c std::unordered_map is also possible.
+
+\code
+using namespace DGtal;
+typedef KhalimskySpaceND<2,int>           KSpace; // ZxZ
+typedef std::map<Cell, CubicalCellData>   Map;    // each cell must store at least a CubicalCellData
```
dcoeurjo: It's more explicit like that but I remember there are  cell map and cell set in KhalimksyND (using rebind mechanisms) ?

src/DGtal/topology/doc/moduleCubicalComplex.dox
```diff
@@ -0,0 +1,328 @@
...
+
+The following programs are related to this documentation:
+cubical-complex-collapse.cpp, cubical-complex-illustrations.cpp,
+testCubicalComplex.cpp,
+
+@section dgtal_ccomplex_sec1  Introduction to cubical complexes
+
+We define a \b cubical \b complex \a C as a collection of cells living
+in some Khalimsky space. Two cells of \a C are \b incident if and only
+if they are incident in the Khalimsky space.
```
dcoeurjo: maybe you should describe what new services/features are provided by the CubicalComplex compared to Khalimsky.

src/DGtal/topology/doc/moduleCubicalComplex.dox
```diff
@@ -0,0 +1,328 @@
...
+job. Again it is in general slightly faster than using
+CubicalComplex::coFaces.
+
+@section dgtal_ccomplex_sec4  Closure, star and link of cells of a complex
+
+You have three methods to compute these "subcomplexes" of a complex \a
+X. In each case, you must give a subcomplex \a S of \a X which defines
+the cells for which you wish to compute the closure, star or link.
+
+@image html cubical-complex-illustrations-S.png "Cubical complex \a X in metallic grey and subcomplex \a S in dark grey."
```
dcoeurjo: metallic->light ?

src/DGtal/topology/doc/moduleCubicalComplex.dox
```diff
@@ -0,0 +1,328 @@
...
+  
+- CubicalComplex::getInteriorAndBoundary: returns both the boundary
+  and the interior of X, faster then computing both separately if you
+  need both.
+
+@section dgtal_ccomplex_sec6 Set operations on complexes
+
+A cubical complex can be seen as a generic container (it is a model of
+boost::Container). Remember though that it is not a model of
+boost::DefaultConstructible, because it must known a cellular grid
```
dcoeurjo: known->know

src/DGtal/topology/doc/moduleCubicalComplex.dox
```diff
@@ -0,0 +1,328 @@
...
+greatest open complex contained in it.
+
+A more difficult topological operation is the \b collapse
+(functions::collapse). A collapse transforms a complex by removing
+cells, but at the end the complex has the same homotopy type as the
+initial complex. The topology is preserved since only free pairs are
+removed. The following snippet shows how to collapse complex X, while
+keeping two cells \a fixed.
+
+@snippet topology/cubical-complex-illustrations.cpp cubical-complex-illustrations-collapse
```
dcoeurjo: in the snippet, you don't describe the ```CC::FIXED``` flag. 
At this point, you haven't yet described what the data attached to cells (CubicalCellData) is used for.

dcoeurjo: ok my bad, there is something line 272. maybe you could detail a little bit this because it is the first usage of a data (flag here) associated with a cell

src/DGtal/topology/doc/moduleCubicalComplex.dox
```diff
@@ -0,0 +1,328 @@
...
+
+As you can see, when collapsing a complex X, you precise:
+- the subcomplex of X that can be collapsed (should be closed,
+  otherwise it is closed).
+- a priority object that tells in which order cell are collapsed. If
```
dcoeurjo: capital letter in items?

src/DGtal/topology/doc/moduleCubicalComplex.dox
```diff
@@ -0,0 +1,328 @@
...
+  cell. You may use CubicalComplex::operator[] to change this value
+  for each cell.
+- various hints to speed up the process.
+
+Example topology/cubical-complex-collapse.cpp shows that it works also in 3D (in fact, it is nD).
+
+@image html cubical-complex-illustrations-collapse.png "Homotopic collapse of \a X, displayed in orange, with fixed points marked in blue."
+@image latex cubical-complex-illustrations-collapse.png "Homotopic collapse of \a X, displayed in orange, with fixed points marked in blue." width=5cm
+
+*/
```
dcoeurjo: (very nice illustrations BTW)

tests/base/testSetFunctions.cpp
```diff
@@ -0,0 +1,326 @@
...
+  Container C1_symdiff_C2 = C1 ^ C2;
+  Container C2_symdiff_C1 = C2 ^ C1;
+  REQUIRE( C1_symdiff_C2.size() == C2_symdiff_C1.size() );
+  REQUIRE( C1_symdiff_C2.size() == ( C1_union_C2 - C1_intersection_C2 ).size() );
+  REQUIRE( C1_symdiff_C2.size() == ( C1_minus_C2 | C2_minus_C1 ).size() );
+
+  // JOL: catch relation operator overloading interacts too much with
+  //      our own set operator overloading. We call the exact
+  //      operations.
+  // std::cout << "C1 D C2 = ";
```
dcoeurjo: maybe you can cleanup this comment since you have removed the overloading on "=="

dcoeurjo:
I still have to dig into implementations of CC but this is really cool.

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: If you prefer, I can use the cell map type of the Khalimsky space as default Map type.

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: done.

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: ok.

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: ok

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: I have written a paragraph is the first section that explains what data are for.

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: Done.

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: Thanks.

tests/base/testSetFunctions.cpp
JacquesOlivierLachaud: cleaned up.

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: Sure.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: Done.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: Done. I don't like it too much because we cannot refer to these constants in the doxygen doc afterwards.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: Done.

src/DGtal/topology/doc/moduleCubicalComplex.dox
JacquesOlivierLachaud: I've done it. I've also fixed a problem in the rebind structure in KhalimskySpaceND. The rebind syntax for a default template parameter is a pain in the ass !

dcoeurjo: yep;) agree..

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1307 @@
...
+     * Get range of equal elements to \a aCell. Because all elements
+     * in a set container are unique, the range returned will contain
+     * a single element at most.
+     *
+     * @param aCell any cell.
+     *
+     * @return the bounds of a range that includes all the elements in
+     * the container that are equivalent to val.
+     */
+    std::pair< ConstIterator, ConstIterator > equal_range( const Cell& aCell ) const;
```
dcoeurjo: equalRange?

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1307 @@
...
+     * Get range of equal elements to \a aCell. Because all elements
+     * in a set container are unique, the range returned will contain
+     * a single element at most.
+     *
+     * @param aCell any cell.
+     *
+     * @return the bounds of a range that includes all the elements in
+     * the container that are equivalent to val.
+     */
+    std::pair< Iterator, Iterator > equal_range( const Cell& aCell );
```
dcoeurjo: same here

src/DGtal/topology/CubicalComplex.ih
```diff
@@ -0,0 +1,1154 @@
...
+  Cells proper_cofaces = myKSpace->uCoFaces( aCell );
+  // NB: (JOL) do not use std::remove_if since predicate is passed by value.
+  typename Cells::iterator first  = proper_cofaces.begin();
+  typename Cells::iterator last   = proper_cofaces.end();
+  typename Cells::iterator result = first;
+  for ( ; first != last; ++first )
+    {
+      if ( belongs( *first ) )
+        {
+#ifdef WITH_C11
```
dcoeurjo: not needed anymore

src/DGtal/topology/CubicalComplex.ih
```diff
@@ -0,0 +1,1154 @@
...
+  Cells proper_faces = myKSpace->uFaces( aCell );
+  // NB: (JOL) do not use std::remove_if since predicate is passed by value.
+  typename Cells::iterator first  = proper_faces.begin();
+  typename Cells::iterator last   = proper_faces.end();
+  typename Cells::iterator result = first;
+  for ( ; first != last; ++first )
+    {
+      if ( belongs( *first ) ) 
+        {
+#ifdef WITH_C11
```
dcoeurjo: same here

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1307 @@
...
+
+    /**
+     * Erase element pointed by iterator \a it.
+     * @param position any iterator on a valid cell.
+     */
+    void erase( Iterator position );
```
dcoeurjo: can't the iterator be const?

src/DGtal/topology/CubicalComplex.h
dcoeurjo: cleanup the other ones?

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1307 @@
...
+    /**
+    * Clears the cubical complex, which becomes empty.
+    */
+    void clear();
+
+    /**
+    * Clears all cells of dimension \a d of the cubical complex.
+    * @param d the dimension of cell \a aCell.
+    */
+    void clear( Dimension d );
```
dcoeurjo: const Dimension

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1307 @@
...
+
+    /**
+    * Fills the data of every cell of dimension \a d this cubical
+    * complex, which becomes \a data. Default value resets the data to
+    * zero.
+    *
+    * @param d the dimension of cell \a aCell.
+    * @param data any data.
+    */
+    void fillData( Dimension d, Data data = Data() );
```
dcoeurjo: const here too (and elsewhere with Dimension)

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1307 @@
...
+    * @return 'true' if and only if \a aCell belongs to this complex.
+    */
+    bool belongs( const Cell& aCell ) const;
+
+    /**
+    * @param d the dimension of cell \a aCell.
+    * @param aCell any cell valid in the Khalimsky space associated to the complex.
+    * @return 'true' if and only if \a aCell belongs to this complex.
+    */
+    bool belongs( Dimension d, const Cell& aCell ) const;
```
dcoeurjo: const Dimension

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1307 @@
...
+    * default.
+    *
+    * @param d the dimension of all cells in the range [it,itE).
+    * @param it an iterator pointing at the beginning of a range of (arbitrary) cells.
+    * @param itE an iterator pointing after the end of a range of (arbitrary) cells.
+    * @param data any value.
+    * @tparam CellConstIterator any model of a forward const iterator on Cell.
+    */
+    template <typename CellConstIterator>
+    void insertCells( Dimension d, CellConstIterator it, CellConstIterator itE, const Data& data = Data() );
```
dcoeurjo: const on iterators ?

src/DGtal/topology/CubicalComplex.h
```diff
@@ -0,0 +1,1307 @@
...
+    */
+    CubicalComplex ( const CubicalComplex & other );
+    
+    /**
+    * Constructor a complex from a digital set.
+    * @param set - a digital set from which to create a complex. 
+    * Set has to be of the same dimension as a Khalimsky space.
+    */
+    template < typename TDigitalSet >
+    void construct ( const TDigitalSet & set );
```
dcoeurjo: As for DEC, I would suggest to create a separate CubicalComplexHelper which would contain functions to create a CubicalComplex from a digital set, from a digital surface, from a GridCurve...

dcoeurjo:
otherwise it fine for me.. thanks. 
(please tell me for the helper class)

dcoeurjo:
for the digital surface case, I see your point in the documentation. If I've understood correctly, one can easily construct a static cubical complex from a digital surface (remove signs on surfels + closure to add lower dimensional cells) but that all (no facade nor light CubicalComplex since CC is a container), am I right?

dcoeurjo:
oops sorry I've closed this PR, reopening it...

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: This is to satisfy STL concepts for associative containers, so I think it is better to keep them lower case.

src/DGtal/topology/CubicalComplex.ih
JacquesOlivierLachaud: itou

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: Ok, this is true only since c++11, before it was iterator.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: There is a "Guru of the week" paper by Herb Sutter on this "const" for pass by value parameter, and he advises against this practice. http://www.gotw.ca/gotw/006.htm
I know some people do it.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: Same remark.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: This was added by @copyme .
I propose to keep it here and to do a helper with more construct in a separate PR.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: I prefer not, because CSTLAssociativeContainer is not "standard" while the others are. They should be put back when they are fixed. I think the comment in the file is not a problem and explains why we check the concept like that.

JacquesOlivierLachaud:
Should be ready now.

src/DGtal/topology/CubicalComplex.h
dcoeurjo: Beside compiler optimization, setting the const is more informative for the user (and I don't see where could be the flaw here from the gotw link).

src/DGtal/topology/CubicalComplex.h
dcoeurjo: The text comment is important because you clarify the problem. I'm afraid that the commented code will stay indefinitely.

src/DGtal/topology/CubicalComplex.h
JacquesOlivierLachaud: It is passed by value, so there is no difference for the user, and more time consuming for him to add a const. Some people do advise to add the const only in the definition of the method (not the declaration). Even so, it is so easy for a compiler to detect that a variable is only read that I do not see what prevents him from optimizing it.

dcoeurjo: I see.

src/DGtal/topology/CubicalComplex.h
dcoeurjo: //JOL: unordered_set and unordered_map do not respect boost::AssociativeContainer concept (boost issue),
      //so we only use DGtal implementation of this concept check.

dcoeurjo:
Perfetct... thanks. merging

PR #1080, opened 2015-11-30
Requiring CPP11 by default in DGtal

dcoeurjo:
This PR:

* remove some ```#ifdef WITH_C11``` switches and update the code accordingly
* change some deprecated ```boost::array``` to ```std::array```
* various cpp11 related fixes


dcoeurjo:
I'm still working on it, it just wanted to notify the devs.

rolanddenis:
:+1:

dcoeurjo:
@rolanddenis I have an issue with catch on travis (see https://gist.github.com/dcoeurjo/888095e6f971e2d25267)... any idea?

rolanddenis:
It seems that `gcc` doesn't support `std::underlying_type` until version 4.7 (see eg http://stackoverflow.com/questions/7931378/how-do-i-get-the-fundamental-type-of-an-enum)
A fix could be to define `CATCH_CONFIG_CPP11_NO_IS_ENUM` when a previous `gcc` version is used.

dcoeurjo:
@JacquesOlivierLachaud : std::binder1st is deprecated in c++11 (see LocalEstimatorFromSurfelFunctorAdapter.h, line 145) and an error is raised on travis. 

Coud you please have a look to the code ? It is unclear for me how to fix it with C++11 stuffs (better solution than redefining a "binder" inner struct)

rolanddenis:
A way to solve this is to replace
```typedef std::binder1st<Metric> MetricToPoint;```
by
```typedef std::function< typename Metric::Value ( typename Metric::Point ) > MetricToPoint;```
including `<functional>` header.

In `LocalEstimatorFromSurfelFunctorAdapter.ih`, line 170, the construction of `metricToPoint` becomes:
```MetricToPoint metricToPoint = std::bind( *myMetric, myEmbedder( *it ), std::placeholders::_1 );```
(adding constness to `metricToPoint` and `vfunctor` is possible).

It is also possible to directly construct `vfunctor` from a lambda and thus avoiding declaring those private typedef.

Note: the documentation talks about a `CMetric` concept, is it not rather `CMetricSpace` ?

**Comment by dcoeurjo** [comment]
Thanks  @rolanddenis !

**Comment by dcoeurjo** [comment]
(Yeah.. CMetric -> CMetricSpace, fixed now)

**Comment by dcoeurjo** [comment]
Allright, this PR is ready for review.

**Comment by kerautret** [comment]
I look it and also with DGtalTools

**Comment by rolanddenis** [comment]
The "WITH_C11" value is still displayed and "false".
See https://github.com/dcoeurjo/DGtal/blob/cpp11/cmake/CheckDGtalOptionalDependencies.cmake#L39

**Comment by dcoeurjo** [comment]
you're right. I'll clean this up

File: cmake/CheckDGtalDependencies.cmake [hunk 1]
```diff
@@ -20,6 +20,57 @@ if ( Boost_FOUND )
...
+    SET(C11_FEATURES "${C11_FEATURES} std::array")
+  ENDIF()
+  IF (CPP11_RREF_MOVE)
+    SET(C11_RREF_MOVE 1)
+    SET(C11_FEATURES "${C11_FEATURES} std::move rvalue-reference(&&)")
+  ENDIF()
+  MESSAGE(STATUS "Supported c++11 features: [${C11_FEATURES} ]")
+  ADD_DEFINITIONS("-DWITH_C11 ")
+ELSE()
+  MESSAGE(FATAL_ERROR "Your compiler does not support any c++11 feature. Please specify another C++ compiler of disable this WITH_C11 option.")
```
kerautret: since we can no more use it without c++11 we can update "or disable this WITH_C11 option"

kerautret:
looks fine for me, tested on different compilators, and it looks also to fix the DGtalTools "rand" related bug. ;) !
Just correct also the message that appears on non cpp11 compilers.
And also the main documentation we should precise it: (in  moduleBuildDGtal.dox) ?
(and also here  http://dgtal.org/download/)

dcoeurjo:
yep, you're right.

dcoeurjo:
I've also changed the dgtal.org web page. 
Ready for a merge ?

kerautret:
yes thanks, perfect ! I have also the fix in DGtalTools, merging..

dcoeurjo:
Thanks... Please have a look to #1090 for the boost::random cleanup.

PR #1084, opened 2015-12-07
Adding raw I/O capabilities for non integral types and signed integers.

rolanddenis:
E.g. works with `double` or `int` using `exportRaw` and `importRaw` methods.

Note: in the writer, `put` accept a `std::ostream::char_type` which is, by default, a `char`. So, I don't understand why the conversion from an `unsigned char` don't not lead to a bug ...


dcoeurjo:
Tricky bugfix, thanks. 
Could you please add a small unit test (eg to check that no warning are raised on Travis bots when using the class) ?

rolanddenis:
Done. I have added writer test in the reader test file.

tests/io/readers/testRawReader.cpp
```diff
@@ -46,77 +51,288 @@ using namespace std;
...
+  testWriteAndRead<2, double, RawIO>( 1.23456789 );
+}
+
+TEST_CASE( "Checking writing & reading double in 3D with generic IO", "[reader][writer][3D][raw][double]" )
+{
+  testWriteAndRead<3, double, RawIO>( 1.23456789 );
+}
+
+/*
+TEST_CASE( "Checking writing & reading double in 4D with generic IO", "[reader][writer][4D][raw][double]" )
```
dcoeurjo: code cleanup ?

dcoeurjo:
thanks for the test. I only have a minor comment

rolanddenis:
Done ;)

dcoeurjo:
Thanks  
Oops I forgot something: could you please update the  module documentation page ?

rolanddenis:
Added.

dcoeurjo:
In I meant ```moduleIO.dox``` page  :)

rolanddenis:
Haa ok. Should I also update GenericReader and GenericWriter to allow I/O on any type ? Currently, only 8bits and 32bits raw file are supported.

dcoeurjo:
I don't know for GenericWriter/Reader since the extension is used to switch between formats (with implicit cast if the Image value type is not the same).

rolanddenis:
It seems that the extension is always "raw" and that the value type if deduced from the Image::Value type. I'll give it a try ;)

rolanddenis:
Done. I have also fixed some bugs in GenericReader/Writer. By the way, those files are not very understandable and it is not clear whether the same ( file format, value type ) pairs are supported by both classes.

With `C++11`, it is possible to simplify this by using functor return or parameter type to set the file value type.

dcoeurjo:
Thanks a lot.
Merging

PR #1086, opened 2015-12-10
Periodic Khalimsky space and pre-Khalimsky space.

rolanddenis:
## Introduction
This PR adds the possibility to specify a different closure type for each dimension, and to define periodic dimensions. In addition, some bugs in `KhalimskySpaceND` have been fixed.

To ensure that the cells coordinates keep valid (ie inside the boundaries, even for periodic dimensions), many checks have been added in `KhalimskySpaceND` methods to enforce valid manipulation of the cells.

Therefore, in order to keep the ability of low-level coordinate modification, the new class `KhalimskyPreSpaceND` model an unbounded cellular grid space that manipulates so-called *pre-cell*. All his methods are static and allow fast and unchecked read/write of the pre-cells.

## Pre-Space
Even if the conversion cell to pre-cell is allowed, the reverse conversion is only possible through dedicated methods of `KhalimskySpaceND`.

`CPreCellularGridSpace` is a new concept for `KhalimskyPreSpaceND` and `CCellularGridSpace` now depends on `CPreCellularGridSpace` plus few methods related to the bounds.

## Periodic Khalimsky space
Along a periodic dimension, it is always possible to translate a cell without going out of space (`(u|s)IsInside` returns always true and `IsMin` & `IsMax` return always false). The Khalimsky coordinates returned by `(u|s)KCoords` are always between `2*myLower[k]` (like a closed dimension) and `2*myUpper[k]+1` (like an open dimension) where `myLower` and `myUpper` are the digital bounds given in the `init` method. `(u|s)GetMin`, `(u|s)GetMax`, `(u|s)First` and `(u|s)Last` return cells based on the reference domain bounds (like `(u|s)KCoords`).

Thus, storage classes based on a periodic Khalimsky space will only see cells in the reference domain (unless for cells whose coordinates were modified by hand, without using `KhalimskySpaceND` methods). In particular, the cubical complex PR #1079 is compatible with periodic Khalimsky space, at least for the collapsing function.

The previous interface is kept unchanged (in particular the `init` method with only a bool to specify closed space) and thus, current code using `KhalimskySpaceND` should work like before.

When using different closure type per dimension, or for periodic spaces, the method `isSpaceClosed` returns true if every dimension is closed (non-periodic) or periodic. Thus, if `isSpaceClosed` returns false, it doesn't mean that the space is open. 

## Checklist
- [x] Documenting `KhalimskySpaceND`, especially the behaviour of each method with periodic dimension.
- [x] Modifying `CCellularGridSpaceND` ?
- [x] ~~Adding a template parameter to optimize the code when periodicity is not needed (the current performance impact must be tested).~~
- [x] Less coordinates verification for periodic dimension in some methods ?
- [x] Checking compatibility of DGtal with periodic Khalimsky spaces (syntax checking):
  * [x] `SurfelNeighborhood` (comparison with `KhalimskySpaceND::min` & `KhalimskySpaceND::max`),
  * [x] `SurfelAdjacency`,
  * [x] `ExplicitDigitalSurface`,
  * [x] `Surfaces`:
     * [x] ~~`extractAllPointContours4C` (direct coordinates manipulation),~~ ( as an issue )
     * [x] `sWriteBoundary` (direct coordinates manipulation),
     * [x] `uFillInterior` (comparison with `uGetMax`),
     * [x] others.
- [x] Testing DGtalTools tools containing `KSpace` (from `grep -rl 'KSpace' *`)
- [x] Testing DGtal examples that contains `KhalimskySpace` or `KSpace` and comparing it to the master.
- [x] ~~When possible, make classes use `KhalimskyPreSpace` as default Khalimsky space and/or weaken concept check (to CPreCellularGridSpace).~~ ( as an issue )

## Disclaimer
Adding periodic dimensions implies to guarantee validity of the cells (to avoid too high extra computational cost of `KhalimskySpaceND`methods). Therefore, it was needed to deeply refactor this class and that explains this huge PR...
The positive part is that highlights suspicious usages of `KhalimskySpaceND` (actually, some bugs was discovered) and features strong foundations for future classes that rely on bounded spaces.


JacquesOlivierLachaud:
It would be nice also to document the user guide to KhalimskySpaceND accordingly (to add to todo list).

rolanddenis:
I have checked some classes to see if the syntax is compatible with periodic Khalimsky Space. Not compatible syntax is e.g.:
- modifying "by hand" cells coordinate without using afterwards the `KhakimskySpaceND` methods that guarantee the correctness of the coordinates (like `(u|s)Cell`, `(u|s)KCoords`, etc).
- comparing cells coordinates with the result of `KhalimskySpaceND::(min|max)`, `(u|s)Get(Min|Max)`, `(u|s)(First|Last)` instead of using dedicated methods `(u|s)Is(Min|Max)`.

The checked files and methods are listed in the todo list. Some methods of `Surfaces` probably need to be fixed.

dcoeurjo:
@JacquesOlivierLachaud could you please review this PR ?

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -406,7 +417,19 @@ namespace DGtal
     template <typename Value> struct SurfelMap {
       typedef std::map<SCell,Value> Type;
     };
+
+    /// Boundaries closure type
+    enum Closure
+      {
+        closed,   ///< The space is closed and non-periodic.
```
JacquesOlivierLachaud: I prefer constants to be uppercase.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -438,16 +461,53 @@ namespace DGtal
...
+               const Point & upper,
+               bool isClosed );
+    
+    /**
+     * Specifies the upper and lower bounds for the maximal cells in
+     * this space.
+     *
+     * @param lower the lowest point in this space (digital coords)
+     * @param upper the upper point in this space (digital coords)
+     * @param closure \a closed, \a open or \a periodic if this space is resp. closed (and non-periodic),
```
JacquesOlivierLachaud: uppercase for constants.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -438,16 +461,53 @@ namespace DGtal
...
-               bool closed );
+               Closure closure );
+
+    /**
+     * Specifies the upper and lower bounds for the maximal cells in
+     * this space.
+     *
+     * @param lower the lowest point in this space (digital coords)
+     * @param upper the upper point in this space (digital coords)
+     * @param closure an array of \a closed, \a open or \a periodic if this space is resp. closed (and non-periodic),
```
JacquesOlivierLachaud: Same as above. "in the corresponding..." => "along the corresponding"

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -475,19 +535,60 @@ namespace DGtal
...
     /**
        @return the upper bound for cells in this space.
+       @todo doc for periodic
     */
     const Cell & upperCell() const;
 
+    /// @}
+
+    // ----------------------- Closure type query --------------------------
+    /** @name Closure type query
```
JacquesOlivierLachaud: Nice feature !

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -509,7 +610,7 @@ namespace DGtal
      * @return the cell having the topology of [c] and the given
      * digital coordinates [p].
      */
-    Cell uCell( const Point & p, const Cell & c ) const;
+    Cell uCell( Point p, const Cell & c ) const;
```
JacquesOlivierLachaud: Why passing Point by value ? It could be costly for spaces of large dimension.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -531,7 +632,7 @@ namespace DGtal
      * @return the cell having the topology and sign of [c] and the given
      * digital coordinates [p].
      */
-    SCell sCell( const Point & p, const SCell & c ) const;
+    SCell sCell( Point p, const SCell & c ) const;
```
JacquesOlivierLachaud: Same as above.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -541,7 +642,7 @@ namespace DGtal
      *
      * @return the spel having the given digital coordinates [p].
      */
-    Cell uSpel( const Point & p ) const;
+    Cell uSpel( Point p ) const;
```
JacquesOlivierLachaud: Same as above.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -552,7 +653,7 @@ namespace DGtal
      *
      * @return the signed spel having the given digital coordinates [p].
      */
-    SCell sSpel( const Point & p, Sign sign = POS ) const;
+    SCell sSpel( Point p, Sign sign = POS ) const;
```
JacquesOlivierLachaud: Same as above.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -562,7 +663,7 @@ namespace DGtal
      *
      * @return the pointel having the given digital coordinates [p].
      */
-    Cell uPointel( const Point & p ) const;
+    Cell uPointel( Point p ) const;
```
JacquesOlivierLachaud: Same as above.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -573,10 +674,14 @@ namespace DGtal
      *
      * @return the signed pointel having the given digital coordinates [p].
      */
-    SCell sPointel( const Point & p, Sign sign = POS ) const;
+    SCell sPointel( Point p, Sign sign = POS ) const;
```
JacquesOlivierLachaud: Same as above.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -645,7 +755,7 @@ namespace DGtal
      * @param k any valid dimension.
      * @param i an integer coordinate within the space.
      */
-    void uSetKCoord( Cell & c, Dimension k, const Integer & i ) const;
+    void uSetKCoord( Cell & c, Dimension k, Integer i ) const;
```
JacquesOlivierLachaud: Better yes, because int is small.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -912,7 +1052,7 @@ namespace DGtal
        @return the same element as [p] except for the incremented
        coordinate [k].
     */
-    Cell uGetIncr( const Cell & p, Dimension k ) const;
+    Cell uGetIncr( Cell p, Dimension k ) const;
```
JacquesOlivierLachaud: Same question as above.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -943,7 +1083,7 @@ namespace DGtal
        @return the cell similar to [p] but with the maximum allowed
        [k]-coordinate.
     */
-    Cell uGetMax( const Cell & p, Dimension k ) const;
+    Cell uGetMax( Cell p, Dimension k ) const;
```
JacquesOlivierLachaud: Same question as above.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -985,7 +1125,7 @@ namespace DGtal
        @return the same element as [p] except for a coordinate [k]
        incremented with x.
     */
-    Cell uGetAdd( const Cell & p, Dimension k, const Integer & x ) const;
+    Cell uGetAdd( Cell p, Dimension k, const Integer & x ) const;
```
JacquesOlivierLachaud: You could pass x by value as above.

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -996,7 +1136,7 @@ namespace DGtal
        @return the same element as [p] except for a coordinate [k]
        decremented with x.
     */
-    Cell uGetSub( const Cell & p, Dimension k, const Integer & x ) const;
+    Cell uGetSub( Cell p, Dimension k, const Integer & x ) const;
```
JacquesOlivierLachaud: You could pass x by value as above.

src/DGtal/topology/KhalimskySpaceND.h
rolanddenis: Because the point was anyway copied and modified in the definition. But maybe the previous syntax is better ?

JacquesOlivierLachaud: It was just wondering. If it is copied inside, then your syntax is better.

src/DGtal/topology/KhalimskySpaceND.h
rolanddenis: In fact, I did that because `i` is modified in the definition :)

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -438,62 +532,207 @@ namespace DGtal
...
+     * @{
+     */
   public:
 
     /**
+     * From a cell, returns the same cell with appropriate
+     * Khalimsky coordinates when the space has periodic
+     * dimension.
+     *
+     * This is only useful for convertion between non-periodic
```
JacquesOlivierLachaud: conversion

src/DGtal/topology/KhalimskySpaceND.h
```diff
@@ -503,18 +742,40 @@ namespace DGtal
...
      */
-    Cell uCell( const Point & p, const Cell & c ) const;
+    Cell uCell( Point p, const Cell & c ) const;
+
+    /**
+     * From a signed cell, returns the same signed cell with
+     * appropriate Khalimsky coordinates when the space has
+     * periodic dimension.
+     *
+     * This is only useful for convertion between non-periodic
```
JacquesOlivierLachaud: conversion

JacquesOlivierLachaud:
Should you not add asserts in methods uCell, uSpel, uPointel and so on, which creates cells from Point coordinate ?

rolanddenis:
Yes, it is the next step :) I am currently adding the pre and post-conditions to each method.

JacquesOlivierLachaud:
Ok.

JacquesOlivierLachaud:
Perhaps call uSetCoord in uSetCoords instead of copy/paste the code (this is true also for uSetKCoord, and signed variants).

JacquesOlivierLachaud:
In uIsInside( const Cell& c ), should you note call isEveryDimensionPeriodic instead of isAnyDimensionPeriodic ?

rolanddenis:
Yes, I've seen the mistake and fix it in the last commits.

JacquesOlivierLachaud:
Sorry, I read the file since I cannot use github.

rolanddenis:
Should we split the `ih` file to allow his visualization in Github ? (But we will lose diff with previous versions)

JacquesOlivierLachaud:
No, it is ok. We will manage like this

JacquesOlivierLachaud:
In (u|s)Next, if you assume that bounds are valid, is it not enough to add 2 to the k-th coordinate ?

rolanddenis:
Yes and no. Yes because I have added the coordinate correction without reading the code.
Finally no because we can use it to span a range that crossed the "fundamental" domain.

JacquesOlivierLachaud:
Smart answer, I think it can help for defining bounding boxes "accross" periodic faces.

rolanddenis:
Nice, smart example of usage :D

JacquesOlivierLachaud:
:)

JacquesOlivierLachaud:
Looks fine to me. I wait for your further updates to complete the todo list.

JacquesOlivierLachaud:
In the todo list, you also need to change the ChangeLog.

rolanddenis:
Should uFaces and related methods check for returned cell validity ? 
i.e. adding only cells that are in the space ?

JacquesOlivierLachaud:
You're right, they should output only what belongs to the space.

rolanddenis:
The doc is finished, I will fix `uFaces` (and related methods) and add assertions related to pre-conditions.

rolanddenis:
The added assertions doesn't seem to slow down the tests : around 27s to do the whole tests with either this PR or the master branch of DGtal.

rolanddenis:
@JacquesOlivierLachaud I have modified the behavior of uCell and sCell. The KhalimskySpaceND class is now ready for review.

JacquesOlivierLachaud:
In KhalimskySpaceND.h, line 407, do you still need the warning ?

rolanddenis:
I'm not sure that every algorithm works as expected with Khalimsky space that have different closure per dimension, or even periodic dimension. I am checking the syntax in some files but I do no go further. What do you think ?

JacquesOlivierLachaud:
We can leave it as is for this PR. We will remove it if it appears that most DGtal algorithms work fine.

rolanddenis:
Ok.

JacquesOlivierLachaud:
For travis error in testIntegralInvariantVolumeEstimator, @jlevallois could you have a look ? Roland has added many asserts in this new version of KhalimskySpaceND. In particular, it can handle periodic spaces but also requires cell to have valid coordinates (within bounds). The assertion error is related to the fact that the shifted cell is not a valid cell of the space. We should either ignore the cell (so use sIsValid), or project it onto the space bounds (with sCell). What do you think ?

JacquesOlivierLachaud:
I've gone through the code and it seems fine to me except for the error on testIntegralInvariantVolumeEstimator (@jlevallois). I think also it would be a good idea to update testKhalimskySpaceND so that  it tests the different kinds of space topology.

jlevallois:
@JacquesOlivierLachaud @rolanddenis It should be ok now, please check my PR https://github.com/rolanddenis/DGtal/pull/1 . Sorry for the delay.

dcoeurjo:
@rolanddenis, @JacquesOlivierLachaud, can you please have a look to @jlevallois's PR ?

dcoeurjo:
kind you please have a look to @jlevallois's edits ? I would like to merge this branch and prepare the 0.9.1

rolanddenis:
I'm currently investigating some issues raised by @jlevallois.

dcoeurjo:
Can you please give me feedbacks on this PR ? I would like to release 0.9.1with this feature.
thx

rolanddenis:
We will continue here the discussion about issues raised by @jlevallois in rolanddenis#1.

dcoeurjo:
All right, I've discussed with both @jlevallois and @rolanddenis today and we decided to postpone this PR after the 0.9.1 release. 
There are strange side effects that need to be investigated.

rolanddenis:
@jlevallois I have tested the current code with a slight modification : at this time, `uCell` and `sCell` modify given coordinates so that the returned cell lies into the Khalimsky Space. For example, if you want to create a cell with Khalimsky coordinates (-3, 2) in a Khalimsky space of lower Khalimsky bound (0, 0), then `uCell` actually returns a cell with coordinates (0,2). It was added to allow some usages seen in DGtal source code (like calling `uFirst` with an invalid cell of wanted dimension).

If I disable this behaviour and remove all ASSERT (to authorize usage of out-of-bounds cells), I obtain the following visualization with
```./3dCurvatureViewer -i bunny64.vol -r 15 -u 255```
bunny
that corresponds to your reference result.

So, it seems that there is some invalid cells created in the current code. I will think about modifying this behaviour to allow better detection of those issues.

rolanddenis:
@dcoeurjo I think it is not related to periodic Khalimsky space but to the fact that I have added consistency checks (and some correction behaviours) to ensure correct usages.

JacquesOlivierLachaud:
@jlevallois Do you still have problems with this PR for II curvature ?

dcoeurjo:
@rolanddenis, @JacquesOlivierLachaud @jlevallois, is this PR mergable ? Would it possible to have a roadmap/todolist to follow what needs to be done ? 
thx

jlevallois:
Seems good for the Integral Invariant part. Thanks @rolanddenis for the modification!

JacquesOlivierLachaud:
@rolanddenis Do you think it is ready now ?

dcoeurjo:
@rolanddenis @JacquesOlivierLachaud @jlevallois , can this PR be reviewed ? what is missing ?

rolanddenis:
I have to update the module documentation page on cellular grid space. In fact, there is probably more updates needed in DGtal but I suggest to add those as issues for later...

rolanddenis:
Done :octocat:

JacquesOlivierLachaud:
Looks great to me. Just 3 remarks:
- you should update the dot graph of packageTopologyConcepts1.dox (miss KhalimskyPre...)
- perhaps myCoordinates should be renamed coordinates in preCell ? It is solely if it does not require a lot of changes elsewhere.
- add yourself as author in moduleCellularTopology (author and doc author), so that we can blame you afterwards ;-P

rolanddenis:
I also see a file named `packageTopologyConcepts.dox` that seems to be an old mix of `packageTopologyConcepts1.dox` and `packageTopologyConcepts1.dox`. Should I remove it ?

JacquesOlivierLachaud:
Yes, you are right. You can remove it.

rolanddenis:
Ok. I'll also rename `myPositive` in `positive`.

rolanddenis:
Done :octocat:

JacquesOlivierLachaud:
Just fix the conflicts (in packageTopologyConcepts1.dox and KhalimskySpaceND.h) and we can merge.

JacquesOlivierLachaud:
At least, all DGtalTools compile with these modifications. I have checked a few tools and they work.

rolanddenis:
Yes, I have already checked DGtalTools (see the end of this PR's todolist). Actually, this PR helps me to found some bugs in DGtalTools (see https://github.com/DGtal-team/DGtalTools/pull/256) and I have some fixes to pull to DGtalTools now.

dcoeurjo:
Actually, Travis now tries to compile the DGtaltools for each PR (see Travis.Yml).

dcoeurjo:
Thanks everyone for this nice PR

rolanddenis:
Finally, after 6 months :sweat_smile:

PR #1087, opened 2015-12-11
Save light position

kerautret:
overload of some methods in orde to save light source position when the Viewer3D state is saved.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -33,6 +33,11 @@
    (Roland Denis, [#1060](https://github.com/DGtal-team/DGtal/pull/1060))
 
 - *IO*
+
+  - In the Viewer3D, the light source position is now saved in the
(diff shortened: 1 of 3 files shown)


dcoeurjo:
great.. thanks. merging.

dcoeurjo:
Just a question: is ```.qglviewer``` always created when the viewer is on ?

dcoeurjo:
I would prefer to have an optional behavior to avoid having bunch of .qglviewer file on my folders..

kerautret:
yes the .qglviewer is always created by default from the QGLviewer. We can probably disable it, and since now we can load/save it from the keyboard it should be ok.

dcoeurjo:
I would prefer to have the file generated only when a save position/light is performed (personally, I never use this feature;))

PR #1088, opened 2015-12-12
add optional .qglviewer.xml saving

kerautret:
and set to false by default.


dcoeurjo:
Great. Maybe you can add setter/getter methods to change the variable state without recompiling (or add a key shortcut in the viewer)

kerautret:
In fact the key_M+SHIFT call directly the save in .qglViewer and flag is myAutoSaveState ;)

kerautret:
is public

dcoeurjo:
Ok fine

PR #1091, opened 2015-12-14
add getFaceCenter method

kerautret:
Tiny PR just to lighten some codes.

### DIFF
--- a/src/DGtal/shapes/Mesh.ih
+++ b/src/DGtal/shapes/Mesh.ih
@@ -257,15 +257,18 @@ DGtal::Mesh<TPoint>::getFace(unsigned int i)
 
 template<typename TPoint>
 inline
-TPoint 
+typename DGtal::Mesh<TPoint>::RealPoint 
 DGtal::Mesh<TPoint>::getFaceBarycenter(unsigned int i) const
 {
-  TPoint c; 
+  DGtal::Mesh<TPoint>::RealPoint c;
   MeshFace aFace = getFace(i);
   for ( auto &j: aFace){
-    c += getVertex(j);
+    TPoint p = getVertex(j);
+    for (typename TPoint::Dimension i = 0; i < TPoint::dimension; i++){
+      c[i] += static_cast<typename RealPoint::Component>(p[i]) ;
+    }
   }
-  return c/static_cast<typename TPoint::Component>(aFace.size());
+  return c/static_cast<typename RealPoint::Component>(aFace.size());
 }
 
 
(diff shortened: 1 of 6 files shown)


src/DGtal/shapes/Mesh.ih
```diff
@@ -256,6 +256,20 @@ DGtal::Mesh<TPoint>::getFace(unsigned int i)
...
 inline
+TPoint 
+DGtal::Mesh<TPoint>::getFaceCenter(unsigned int i) const
+{
+  TPoint c; 
+  MeshFace aFace = getFace(i);
+  for ( auto &j: aFace){
+    c += getVertex(j);
+  }
+  return c/aFace.size();
```
dcoeurjo: implicit cast to int (or size_t) ?

src/DGtal/shapes/Mesh.ih
```diff
@@ -256,6 +256,20 @@ DGtal::Mesh<TPoint>::getFace(unsigned int i)
 
 template<typename TPoint>
 inline
+TPoint 
+DGtal::Mesh<TPoint>::getFaceCenter(unsigned int i) const
```
dcoeurjo: getFaceBarycenter instead ?

kerautret: yes agree

src/DGtal/shapes/Mesh.ih
kerautret: yes ok I add a static_cast

kerautret:
done thanks

src/DGtal/shapes/Mesh.ih
```diff
@@ -256,6 +257,20 @@ DGtal::Mesh<TPoint>::getFace(unsigned int i)
...
 inline
+TPoint 
+DGtal::Mesh<TPoint>::getFaceBarycenter(unsigned int i) const
+{
+  TPoint c; 
+  MeshFace aFace = getFace(i);
+  for ( auto &j: aFace){
+    c += getVertex(j);
+  }
+  return c/static_cast<typename TPoint::Component>(aFace.size());
```
dcoeurjo: well.. yes.. but If the mesh has integer coordinates vertices, the barycenter is a bit crappy since it may not lie on the surface.

dcoeurjo: maybe you should make this clear. Or decide that the output of "getbarycenter" is always a "RealPoint" and thus cast to double the denominator

kerautret: ok for RealPoint it is more logic

dcoeurjo:
Ok fine.
Before merging, could you please fix conflicts with the master ?

kerautret:
yes done ;)

dcoeurjo:
thanks merging

PR #1101, opened 2016-01-08
Shadow boost warnings at build.

phcerdan:
Setting include_directories(SYSTEM ${Boost_INCLUDE_DIRS})

This shadows boost warnings when using/building DGtal. 
https://cmake.org/cmake/help/v3.0/command/include_directories.html


dcoeurjo:
Excellent. Thanks (I was not aware of this flag) 
Merging

Issue #1095, opened 2015-12-28
Install DGtalConfig.cmake in default lib/cmake/DGtal instead of current /share/DGtal/CMake/


phcerdan:
From: https://cmake.org/Wiki/CMake/Tutorials/Packaging
Note: If your project does not already have a <prefix>/lib/Foo*/ directory you may prefer to put the package file in <prefix>/lib/cmake/Foo*/ to keep the lib directory clean. However, CMake 2.6.2 and lower do not search there. CMake 2.6.3 and above do.

dcoeurjo:
I agree. thanks for the report.

PR #1103, opened 2016-01-09
Fixing cmake config file install

dcoeurjo:
Fixing #1095


kerautret:
I just checked the PR but for the install copy always the file here:
 /usr/local/share/DGtalConfig.cmake

dcoeurjo:
are you sure to have removed your cmake cache ?

kerautret:
oops you are right:
/usr/local/lib/DGtal/DGtalConfig.cmake
so looks fine if I am not wrong (since we don't suppose  cmake > 2.6.3 )

dcoeurjo:
I don't understand what you mean with cmake>2.6.3?

kerautret:
as @phcerdan mention, CMake 2.6.3 and above looks in /lib/cmake/Foo/ but I have checked in the CMakeLists.txt we have cmake_minimum_required (VERSION 2.6) so it can keep it ;)

dcoeurjo:
Got it, thanks.

kerautret:
ok merging all get fines also with DGTalTools from user or travis install :)

PR #1117, opened 2016-01-22
Explicit Instantiation of some Z2i/Z3i classes

dcoeurjo:
* cleanup of some cpp included in libDGtal
* explicit instantiation of some Z2i:: and Z3i:: classes
* this PR (slightly) speeds up unit test build


dcoeurjo:
Can someone have a quick look at this PR ?

src/DGtal/geometry/volumes/distance/ExactPredicateLpPowerSeparableMetric.ih
```diff
@@ -311,7 +311,7 @@ DGtal::ExactPredicateLpPowerSeparableMetric<T,2,P>::binarySearchHidden(const Abs
 								       const Abscissa &/*upper*/) const
 {   
   ASSERT(false && "Not Necessary for l_2");
-  
+  return 0;
```
rolanddenis: Should we not better thrown an exception or simply not define this function (only the header) ?

src/DGtal/geometry/volumes/distance/ExactPredicateLpSeparableMetric.ih
```diff
@@ -318,7 +318,7 @@ DGtal::ExactPredicateLpSeparableMetric<T,2,P>::binarySearchHidden(const Abscissa
                                                                   const Abscissa &) const
 {
   ASSERT(false && "Not Necessary for l_2");
-
+  return 0;
```
rolanddenis: Same remark.

src/DGtal/helpers/StdDefs.cpp
```diff
@@ -0,0 +1,68 @@
...
+template class DGtal::KhalimskySpaceND< 2, DGtal::Z2i::Integer > ;
+template class DGtal::HyperRectDomain< DGtal::Z2i::Space > ; 
+template class DGtal::MetricAdjacency< DGtal::Z2i::Space, 1> ;
+template class DGtal::MetricAdjacency< DGtal::Z2i::Space, 2> ;
+template class DGtal::DigitalTopology< DGtal::Z2i::Adj4, DGtal::Z2i::Adj8 > ;
+template class DGtal::DigitalTopology< DGtal::Z2i::Adj8, DGtal::Z2i::Adj4 > ;
+template class DGtal::DigitalSetByAssociativeContainer<DGtal::Z2i::Domain,
+                                                       std::unordered_set< typename DGtal::Z2i::Domain::Point> >;
+//template class DGtal::Object<DGtal::Z2i::DT4_8, DGtal::Z2i::DigitalSet> ;
+//template class DGtal::Object<DGtal::Z2i::DT8_4, DGtal::Z2i::DigitalSet> ;
```
rolanddenis: Why not ?

src/DGtal/helpers/StdDefs.cpp
```diff
@@ -0,0 +1,68 @@
...
+template class DGtal::GridCurve<DGtal::Z2i::K2> ;
+template class DGtal::ExactPredicateLpSeparableMetric<DGtal::Z2i::Space,2> ;
+template class DGtal::ExactPredicateLpSeparableMetric<DGtal::Z2i::Space,1> ;
+template class DGtal::ExactPredicateLpPowerSeparableMetric<DGtal::Z2i::Space,2> ;
+template class DGtal::ExactPredicateLpPowerSeparableMetric<DGtal::Z2i::Space,1> ;
```
rolanddenis: Nice idea ! But should we not do the same for the other `Z2i` typedefs, like `Cell` and `SCell` ?

src/DGtal/helpers/StdDefs.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+//////////////////
+/// Z3i Explicit instanciation
+//////////////////
+template class DGtal::SpaceND<3, DGtal::Z3i::Integer>;
+template class DGtal::PointVector<3, DGtal::Z3i::Integer>;
+template class DGtal::KhalimskySpaceND< 3, DGtal::Z3i::Integer > ;
+template class DGtal::HyperRectDomain< DGtal::Z3i::Space > ; 
+template class DGtal::MetricAdjacency< DGtal::Z3i::Space, 1> ;
+template class DGtal::MetricAdjacency< DGtal::Z3i::Space, 2> ;
```
rolanddenis: ```template class DGtal::MetricAdjacency< DGtal::Z3i::Space, 3>``` ?

src/DGtal/helpers/StdDefs.cpp
```diff
@@ -0,0 +1,68 @@
...
+template class DGtal::DigitalTopology< DGtal::Z3i::Adj6, DGtal::Z3i::Adj26 > ;
+template class DGtal::DigitalTopology< DGtal::Z3i::Adj26, DGtal::Z3i::Adj6 > ;
+template class DGtal::DigitalTopology< DGtal::Z3i::Adj18, DGtal::Z3i::Adj26 > ;
+template class DGtal::DigitalTopology< DGtal::Z3i::Adj26, DGtal::Z3i::Adj18 > ;
+template class DGtal::DigitalSetByAssociativeContainer<DGtal::Z3i::Domain,
+                                                       std::unordered_set< typename DGtal::Z3i::Domain::Point> >;
```
rolanddenis: `Object` and `GridCurve<K3>` ?

src/DGtal/helpers/StdDefs.h
```diff
@@ -183,8 +183,6 @@ namespace DGtal
     static const DT6_26 dt6_26 = DT6_26( adj6, adj26, JORDAN_DT );
     static const DT26_6 dt26_6 = DT26_6( adj26, adj6, JORDAN_DT );
 
-    typedef GridCurve<K3> Curve; 
-
```
rolanddenis: Hoo ok :octocat:

tests/helpers/testStdDefs.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+TEST_CASE( "Testing StdDefs" )
+{
+  
+  //Just for compilation timings.
+  SECTION("Testing domain instanciation (Z2i)")
+    {
+      Z2i::Domain dom(Z2i::Point(0,0), Z2i::Point(16,16));    
+      REQUIRE( dom.isValid() );
+    }
```
rolanddenis: Same for `Z3i` ? :smile:

rolanddenis:
Nice Idea !

src/DGtal/helpers/StdDefs.cpp
dcoeurjo: since Cell ans SCell are included in KSpace,  the associated class would be already compiled, I think.

src/DGtal/helpers/StdDefs.cpp
dcoeurjo: It produces errors at compile time.

src/DGtal/geometry/volumes/distance/ExactPredicateLpPowerSeparableMetric.ih
dcoeurjo: you're right

src/DGtal/helpers/StdDefs.cpp
dcoeurjo: GridCurve asserts at some point that the dimension is 2.
This was an issue I think in the StdDefs.h

src/DGtal/helpers/StdDefs.cpp
rolanddenis: Probably for Cell since it is a member of KSpace but I am not sure for SCell (but I am not confident with explicit instantiation of templates)

src/DGtal/helpers/StdDefs.cpp
dcoeurjo: yup

tests/helpers/testStdDefs.cpp
dcoeurjo: it was just to check build time and the fact that explicit instantiation is used.
but on this simple example, it is roughly the same;)
Should I remove it?

src/DGtal/geometry/volumes/distance/ExactPredicateLpPowerSeparableMetric.ih
rolanddenis: Maybe not the right place to talk about this, but there is a lake of DGtal exceptions, especially with a string member to describe why an exception was raised (e.g. in DGtal::IOException).

src/DGtal/helpers/StdDefs.cpp
dcoeurjo: I remove this 2 lines and open an issue.

src/DGtal/helpers/StdDefs.cpp
dcoeurjo: E.g. I have the feeling that PointVector instantiation is redundant with Space instantiation (and its inner PointVector type).

tests/helpers/testStdDefs.cpp
rolanddenis: Ho I see ! And for the compilation of all tests, is there a speed-up ?
About the test file, since StdDefs.cpp is already compiled, I am not sure that it is needed.

src/DGtal/geometry/volumes/distance/ExactPredicateLpPowerSeparableMetric.ih
dcoeurjo: > Le 22 janv. 2016 à 15:02, rolanddenis <notifications@github.com> a écrit :
> 
> In src/DGtal/geometry/volumes/distance/ExactPredicateLpPowerSeparableMetric.ih <https://github.com/DGtal-team/DGtal/pull/1117#discussion_r50539139>:
> 
> > @@ -311,7 +311,7 @@ DGtal::ExactPredicateLpPowerSeparableMetric<T,2,P>::binarySearchHidden(const Abs
> >  								       const Abscissa &/*upper*/) const
> >  {   
> >    ASSERT(false && "Not Necessary for l_2");
> > -  
> > +  return 0;
> Maybe not the right place to talk about this, but there is a lake of DGtal exceptions, especially with a string member to describe why an exception was raised (e.g. in DGtal::IOException).
> 
> 


I definitely agree… Exception management in DGtal is very limited.

tests/helpers/testStdDefs.cpp
dcoeurjo: I'll redo a timing test but it was not terrific on ```tests/``` maybe because some of them are quite old and do not use Z2i::Z3i...
(i will remove this test)

src/DGtal/helpers/StdDefs.cpp
rolanddenis: I see but it seems that a typedef is not enough to explicitly instantiate a template and since `SpaceND` do not have any member ...
However, `HyperRectDomain` does have point members, so it is probably not needed.

tests/helpers/testStdDefs.cpp
rolanddenis: I am not sure that it is necessary to use Z2i and Z3i to benefit from those explicit instantiations. However, `StdDefs.h` needs probably to be included.

dcoeurjo:
Merging;)

Issue #1025, opened 2015-07-12
line export in Board3D looks strange 


kerautret:
By testing testBoard3D and dgtalBoard3D-line.obj, it looks like that the export is strange:
<img width="791" alt="capture d ecran 2015-07-12 a 16 29 43" src="https://cloud.githubusercontent.com/assets/772865/8638227/87ec1a9e-28b4-11e5-9ad5-a142732d404e.png">

dcoeurjo:
I agree.. it's a duplicate of  #900 I think

PR #1119, opened 2016-01-22
fix issue 1025

kerautret:
(see #1025 ) 
Now we have better export:
capture d ecran 2016-01-23 a 00 47 06


dcoeurjo:
Excellent! 
Thanks, merging

PR #1120, opened 2016-01-23
Set QGLviewer in cache in appveyor...

kerautret:
allow to save some minutes and can be < timeout (60min)


dcoeurjo:
Excellent. Thanks

PR #1126, opened 2016-02-04
indent in example LMST...

kerautret:
Tiny PR, just an indent code correction in doc but which can make wrong when reading..


dcoeurjo:
Thanks

PR #1127, opened 2016-02-05
Contour helper add

kerautret:
add a method to transform a contour into a 8 connected path.


src/DGtal/geometry/helpers/ContourHelper.h
```diff
@@ -75,17 +75,46 @@ class ContourHelper
...
+   **/
+  template<typename TPoint>
+  static 
+  unsigned int 
+  getFreemanCode4C(const TPoint &pt1, const TPoint &pt2);
+  
+
+  /**
+   * Transforms an input contour into an 8 connected contour.
+   * @param[in] itb  begin iterator associated to the input contour.
```
dcoeurjo: what definition of contour do you consider ?
Can't you obtain the same thing with GridCurve ?

src/DGtal/geometry/helpers/ContourHelper.h
```diff
@@ -75,17 +75,46 @@ class ContourHelper
...
+   * @param[in] ite  end iterator associated to the input contour.
+   * @param[out] out   output iterator associated to the resulting contour.
```
dcoeurjo: same here, please define the contour type (what is the type if you dereference the output iterator?)

dcoeurjo:
At this point there are several ways with different types to define a contour on digital 2D objects: GridCurve and DigitalSurface (+ methods in Surfaces.h if I remember correclty). We should be really careful to avoid confusion if we add an new helper on that.

src/DGtal/geometry/helpers/ContourHelper.h
kerautret: I was hoping to use both std::vector<Point> or grid curve but it fails with GridCurve.
Perhaps I can replace with GridCurve

src/DGtal/geometry/helpers/ContourHelper.h
kerautret: the back_inserter fails with grid curve  ;(

kerautret:
> At this point there are several ways with different types to define a contour on digital 2D objects: GridCurve and DigitalSurface (+ methods in Surfaces.h if I remember correclty). We should be really careful to avoid confusion if we add an new helper on that.

In fact here the small helper were defined to filter some open contours which are not extracted from topology definitions but which come from canny filter or similar (so DigitalSurface can not be used).

dcoeurjo:
can you for instance create a 8-connected sequence  from a 4-connected  set of pixels ? Are there assumptions on the input point set  ?

kerautret:
yes it can. It can also work on non integer points but it will have no effect if the input has no 4-connected directions.

dcoeurjo:
Sorry @kerautret for being late on this PR. However, I still don't get it ;)
Could you please describe me again what are the requirements on the input std::vector ?

kerautret:
@dcoeurjo I am well late also ;à)
I follow the discussion and change with a GridCurve for the input (instead initially the InputIterator), I am agree it is more safe since the input should always be 4 connected and integer.

kerautret:
I changed again with InputIterator since GridCurve limits the type of input contour (can not contain 8 connected part)

src/DGtal/geometry/helpers/ContourHelper.h
```diff
@@ -64,7 +65,7 @@ class ContourHelper
   /**
    * Compute the barycenter of the 2D contour.  
    * 
-   * @param aContour the vector containing the 2D contour coordinates.
+   * @param[in] aContour the vector containing the 2D contour coordinates.
```
dcoeurjo: Contour -> a set of points ? (no need to be connected)

dcoeurjo: getBarycenter ?

src/DGtal/geometry/helpers/ContourHelper.h
```diff
@@ -75,16 +76,45 @@ class ContourHelper
...
    *
-   * @param aContour the vector containing the 2D contour coordinates.
+   * @param[in] aContour the vector containing the 2D contour coordinates.
    * @return 'true' if the contour is counter clockwise oriented, 'false' otherwise.
    **/
-
   template <typename TPoint> 
   static
   bool
   isCounterClockWise(const std::vector<TPoint> & aContour);
```
dcoeurjo: from a contour or from a polygonal curve?

src/DGtal/geometry/helpers/ContourHelper.h
```diff
@@ -75,16 +76,45 @@ class ContourHelper
...
+   **/
+  template<typename TPoint>
+  static 
+  unsigned int 
+  getFreemanCode4C(const TPoint &pt1, const TPoint &pt2);
+  
+
+  /**
+   * Transforms an input contour into an 8 connected contour.
+   * The contour is given from a range of points.
```
dcoeurjo: from two iterators on points

dcoeurjo: if the point sequence is not 8 connected, what happens?
This should be described here I think

src/DGtal/geometry/helpers/ContourHelper.ih
```diff
@@ -44,37 +44,100 @@
...
-    ptMean[0]+=aContour.at(i)[0];
-    ptMean[1]+=aContour.at(i)[1];
-  }
+  for(unsigned int i =0; i<aContour.size(); i++)
+    {
+      ptMean[0]+=aContour.at(i)[0];
+      ptMean[1]+=aContour.at(i)[1];
+    }
   ptMean[0] /= aContour.size();
   ptMean[1] /= aContour.size();
```
dcoeurjo: Ok. The barycenter should probably be a RealPoint to avoid rounding decision here.

dcoeurjo:
there was no (point,point)->freeman code method elsewhere ?

src/DGtal/geometry/helpers/ContourHelper.h
kerautret: yes agree I do it

src/DGtal/geometry/helpers/ContourHelper.h
kerautret: both I change with curve

src/DGtal/geometry/helpers/ContourHelper.h
kerautret: if the replacement differs from 4 path no transformation is done, yes I add it

src/DGtal/geometry/helpers/ContourHelper.ih
kerautret: ok done (also given in nd)

kerautret:
in readFromPointsRange given to construct a FC from points range.
I apply a call to this function from the FC initial code (and use the simplified code transform)...

kerautret:
@dcoeurjo thanks for the review, done (and anticipate the changes on DGtalTOols to have Travis OK)

src/DGtal/geometry/curves/FreemanChain.ih
```diff
@@ -678,9 +678,7 @@ void DGtal::FreemanChain<TInteger>::readFromPointsRange( const TConstIterator& i
       {
         
         Point ptSuiv( *it ); 
-        Integer dx = ptSuiv[0] - pt[0]; 
-        Integer dy = ptSuiv[1] - pt[1]; 
-        short number = ( dx != 0 ? (1 - dx) : (2 - dy) );
+        short number = ContourHelper::getFreemanCode4C(pt, ptSuiv);
```
dcoeurjo: I would have preferred to use the method FreemanChain::displacement(dx,dy) instead of an external method.
What do you think ?

src/DGtal/geometry/helpers/ContourHelper.h
```diff
@@ -62,29 +63,63 @@ class ContourHelper
...
+  
 
-   
+  /**
+   * Transforms an input contour into an 8 connected contour.  The
+   * contour is given from two iterators on points.  If the input
+   * sequence contains some points not 4-connected, they are
+   * are ignored (and the transformation continue on the remaining
+   * parts).
+   *
```
dcoeurjo: to understand: if the first point is disconnected and the rest 4-connected, then only the first point is in the output, am I right ?

So maybe the method could be named differently, it returns the longest 8-connected curve in a greedy way.

src/DGtal/geometry/curves/FreemanChain.ih
kerautret: yes sure, I miss it, thanks  ;)

src/DGtal/geometry/helpers/ContourHelper.h
kerautret: no in fact it try to transform all possible sequences

src/DGtal/geometry/curves/FreemanChain.ih
kerautret: in fact displacement is the contrary, it gives the move from the code but I look to write the equivalent.

kerautret:
@dcoeurjo thanks, done by adding the previous   getFreemanCode4C -> FreemanChain::freemanCode4C(dx,dy) and updating comment on pixels2pixels8C

dcoeurjo:
ok thanks :)

dcoeurjo:
merging

PR #1128, opened 2016-02-08
Light viewer custom

kerautret:
Add three new modes for shape rendering (default, metallic and plastic). The rendering can be changed by using the key M. The user can also choose its own rendering with some setter/getter on the opengl
lightning/properties.

capture d ecran 2016-02-08 a 23 14 54
capture d ecran 2016-02-08 a 23 15 00
capture d ecran 2016-02-08 a 23 14 47


dcoeurjo:
Nice!
testBallQuadViewer looks great with this modifier.

PR #1129, opened 2016-02-09
Bug fix: without 'struct', we can't include <DGtal/io/viewers/Viewer3…

akrah:
…D.h> in an external project.

DrawableWithViewer3D is a structure define in "DGtal/base/Common.h" when you compîle DGtal with the WITH_VISU3D_QGLVIEWER option.
I suppose that this can be reproduce just by trying to compile a small code with the "#include <DGtal/io/viewers/Viewer3D.h>" line.

### DIFF
--- a/src/DGtal/io/viewers/CDrawableWithViewer3D.h
+++ b/src/DGtal/io/viewers/CDrawableWithViewer3D.h
@@ -110,7 +110,7 @@ An object x satisfying this concept may then be used as:
   private:
 
     T myT; //! the drawable class
-    DrawableWithViewer3D *myD;
+    struct DrawableWithViewer3D *myD;
 
     std::string myS;
 


dcoeurjo:
Strange issue since DGtaltools (and many others) used the viewer without any problem.

dcoeurjo:
In the doc, I think we mention that you MUST include Comon.h before any other includ. 

Would that solve the problem?

akrah:
No, I tested bu it's the same problem.
I tried to follow all steps of the 3dImageViewer tool, with all the same include directives, but I always have this problem.

akrah:
Finally, considering the definitiion in Common.h, use "struct" seems the good way.
I don't know why 3dImageViewer, for example, has no bug in fact.

dcoeurjo:
Maybe we should correct 3dImageViewer instead of changing the concept. What do you think?

kerautret:
>Maybe we should correct 3dImageViewer instead of changing the concept. What do you think?
"correct Viewer3D" you want say ?

dcoeurjo:
Nope, 3dimageviewer if the Common.h is missing

kerautret:
But 3dImageViewer use the trace for instance which is in Common.h...

akrah:
I think that this is real bug.
You can quickly try this : 

Create the file test.cpp with the following content
```
#include <DGtal/base/Common.h>
#include <DGtal/io/viewers/Viewer3D.h>
int main() { return 0; }
```
Then compile with this commmand
```
g++ test.cpp -o test -fPIC -std=c++11 -DWITH_QT5 -I/usr/include/x86_64-linux-gnu/qt5/QtWidgets -I/usr/include/x86_64-linux-gnu/qt5/ -I/usr/include/x86_64-linux-gnu/qt5/QtCore -I/usr/include/x86_64-linux-gnu/qt5/QtXml -I/usr/include/x86_64-linux-gnu/qt5/QtOpenGL -I/usr/include/x86_64-linux-gnu/qt5/QtGui
```

I obtain a error message in output that begin by
```
In file included from /usr/local/include/DGtal/io/viewers/Viewer3D.ih:53:0,
                 from /usr/local/include/DGtal/io/viewers/Viewer3D.h:1374,
                 from test.cpp:2:
/usr/local/include/DGtal/io/viewers/CDrawableWithViewer3D.h:113:5: error: ‘DrawableWithViewer3D’ does not name a type
     DrawableWithViewer3D *myD;
```

dcoeurjo:
Strange because all dgtaltools with QGLVIEWERS and my tools compile

dcoeurjo:
there is no -DWITH_QGLVIEWER  or VISU.. in your command line ?

dcoeurjo:
``` -DWITH_VISU3D_QGLVIEWER``` ?

**Comment by akrah** [comment]
All dgtaltools compile for me too.
I tried -DWITH_QGLVIEWER and the bug persists.
Perhaps DGtalTools use some options ?
But the current code without a 'struct' seems to me be a bug. And I suppose that my small code prove that, no ?

**Comment by dcoeurjo** [comment]
Please use the cmake script as described in the doc. It would set the variable properly

**Comment by dcoeurjo** [comment]
http://dgtal.org/for-users/

**Comment by dcoeurjo** [comment]
The variable -DWITH_VISU3D_QGLVIEWER must be defined

**Comment by akrah** [comment]
Ok, this code compile with -DWITH_VISU3D_QGLVIEWER, sorry for your time.

I dont't know if you have this somewhere but to include DGtal in a Qt project, you need to add these lines to your .pro : 
```
LIBS += -lDGtal -lQGLViewer
DEFINES += WITH_QT5 WITH_VISU3D_QGLVIEWER
```
With in DEFINES instruction all options used for the DGtal compilation
Maybe can we add this in the documentation ?

dcoeurjo:
There are other dependencies that need special treatment (ITK,...) in DGtal. That is why we provide a DGtalConfig.cmake that does the job for you. 

We could think of generating and maintaining a .PRO script but I have the feeling that we should promote only one system (cmake) that could generate plenty of build solutions than having different ones (qmake, premake, scons, ant....)

dcoeurjo:
(Can I close this issue @akrah  ?)

akrah:
Yes, I understand.
I just think that if you are able to manage all other dependencies (as ITK) individually in your system, this could be a good think to know the list of options you can add to your system (all the WITH_XXX) to be able to use DGtal with your system.
I suppose that just indicate the "WITH list" could be interesting for peoples using any other system than cmake : it was the only thing we were missing.

akrah:
Yes, you can close ! Thanks again !

dcoeurjo:
ok I see... there is the doc page : http://dgtal.org/doc/nightly/moduleHowToUseDGtal-qmake.html

but seems to be deprecated and could be improved.

akrah:
Ok, I can update this doc.
Thanks, this issue can be closed !

dcoeurjo:
closing;)

PR #1130, opened 2016-02-12
Fix vol/pgm/raw reading bug on Win 10

kerautret:
Fix a vol/pgm/raw reading bug occurring on Windows 10 due to the  different interpretations of end of line \r\n on Window versus \n on unix/mac.  Changing reading mode with binary mode instead text mode fix  the issue.
We have now the tools sliceViewer working on Windows 10:
capture d ecran 2016-02-13 a 00 22 21

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -17,8 +17,8 @@
    [#1128](https://github.com/DGtal-team/DGtal/pull/1128))
  - readers: fix a vol/pgm/raw reading bug occurring on Windows 10 due to the
    different interpretations of end of line \r\n on Window versus \n on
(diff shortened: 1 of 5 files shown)


dcoeurjo:
For Vol and Raw, I agree with you (and thanks for having checked on MS ;)). Are you sure it is necessary for PGM (that can be in ascii mode) ?

kerautret:
yes sure, to fix it I started from the PPM reader which was falling on reading pgm image. In fact the reading process is the same (even ascii or not) and the binary mode differs just from some translations on formatting of special characters (see http://www.cplusplus.com/doc/tutorial/files/).

dcoeurjo:
ok thanks for the comment.
merging

PR #1131, opened 2016-02-14
Travis modification to show output on failed tests.

phcerdan:
Pass argument from make test to ctest:
make test ARGS=--output-on-failure
Showing verbose output on failed tests.


phcerdan:
It works in all Travis configurations, it shows verbose output only in failing tests.
I will revert the introduced test failure and squash the extra commits after your review.

dcoeurjo:
Excellent !! I didn't know this make test option. Can you please revert the bug commit ?

Thanks a lot

phcerdan:
Done! The option came from [ctest](https://cmake.org/cmake/help/v3.0/manual/ctest.1.html).

dcoeurjo:
Great;) Many thanks.

PR #1138, opened 2016-03-03
fix bug show() is called at the end of prg

kerautret:
# PR Description
Fix a bug which appears with this case of use:

```
viewer << Z3i::Point(0,0,0);
viewer << MyViewer::updateDisplay;
viewer.show();
return application.exec();
```

instead :
```
viewer.show();
viewer << Z3i::Point(0,0,0);
viewer << MyViewer::updateDisplay;
return application.exec();
```

# Checklist

- [n.a. ] Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).
- [n.a. ] Documentation module page added or updated.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -19,8 +19,8 @@
 
 - *IO Package*
 
- - Viewer3D: Fix a problem when the show() method was called at the of the main
-   program (the list creation was not called).
+ - Viewer3D: Fix a problem when the show() method was called at the end of the
+   main program (the list creation was not called).
    (Bertrand Kerautret [##1138](https://github.com/DGtal-team/DGtal/pull/1138))
  - Viewer3D: add three new modes for shape rendering (default, metallic and
    plastic). The rendering can be changed by using the key M. The user can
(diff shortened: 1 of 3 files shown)


dcoeurjo:
just wondering: can't we force the ```show()``` (when doing updateDisplay for instance) so that the user won't have to call this function ?

kerautret:
thanks looks fine! (I was wondering if we add it also for HueColorMap ?)

kerautret:
calling show() in qt is more to display the window on screen perhaps strange if we hide it (cc @akrah ?).
We could add a flag to check it but however sometimes to do offline screenshot it should be not good.

ChangeLog.md
```diff
@@ -18,6 +18,10 @@
 
 
 - *IO Package*
+
+ - Viewer3D: Fix a problem when the show() method was called at the of the main
```
dcoeurjo: at the END of the main ?

kerautret: yes thanks

dcoeurjo:
thanks merging.

PR #1140, opened 2016-03-08
Updating (Const)Alias and adding ConstAlias where temporaries are not welcome.

rolanddenis:
# PR Description
- Alias and ConstAlias have been updated to raise compilation error when incorrectly constructed (e.g. from rvalue reference).
- Adding ConstAlias in class constructors where temporary parameters lead to invalid objects (when a member is a constant pointer or a constant reference to a parameter).
- Fixing spotted issues (invalid use of temporaries).

For information purpose, concerned classes were spotted using the following commands:
- `grep -nrE 'const\s+[^ &;]*\s*&\s*my[a-zA-Z_0-9]*\s*;' src` for lines like ```const Type & myMember;```
- `grep -nrE 'my[^(]*\s*\(\s*&' src`for lines like ```myMember(&aParam)``` in initializer list.

# Checklist


rolanddenis:
Compilation failure is related to DGtalTools. It seems that `CPP11_*` preprocessor constants was not correctly set when compiling DGtalTools, since everything compiled fine before #1141 .

I will make a PR in DGtalTools :octocat:

dcoeurjo:
ping @elcerdo :smile:, that could interest you

thx a lot @rolanddenis

src/DGtal/base/Alias.h
```diff
@@ -199,61 +199,67 @@ user forward an Alias<T> parameter.
...
 
     /**
-       Constructor from const reference to an instance of T. Invalid.
+       Constructor from const reference to an instance of T. Deleted.
+
+       Aliasing a const-ref is an error. Consider ConstAlias instead.
     */
-    inline Alias( const T& ) : myParam( CONST_LEFT_VALUE_REF ), myPtr( 0 )
-    { ASSERT(( false && "[Alias::Alias( const T& )] Aliasing a const-ref is an error. Consider ConstAlias instead." )); }
+    Alias( const T& ) = delete;
```
dcoeurjo: hein? not sure to clearly see what this does :smile:

dcoeurjo:
excellent PR.
Would it be possible to update the ConstAlias/Alias documentation and ```moduleCloneAndReference``` to describe the best practice and why the user may have a  warning when using rvalue ref?

src/DGtal/base/Alias.h
rolanddenis: We could just not declare those constructors in order to raise a compilation error when used (instead of an error at run-time in Debug mode) but declaring it with `= delete` makes sure that no other constructor will be chosen with const-ref and makes this choice more clear (and documented).

If the question is about the `delete` keyword, it is a new usage in `C++11` and it allows to explicitly say that a function or a member cannot be used (is deleted). For example, it replaces more efficiently the use of defined (but not implemented) private members (like when a class is not copyable).

The other new associated keyword is `default` to explicitly say you want the default behavior for constructor, destructor and assignment/move operators.

rolanddenis: By the way, I was thinking about updating the templates in DGtalScripts to use those keywords and to allow move operations (in fact, declaring a copy operation prevents the compiler to automatically generate move operations).

rolanddenis:
The table http://dgtal.org/doc/nightly/moduleCloneAndReference.html#moduleCloneAndReference_sec21 actually describes the recommenced practices. Does it need update ?

Concerning the compilation error (no warnings here) when using deleted methods of `Alias` of `ConstAlias`, the `= delete` is visible in the member list, at the top of the class documentation, and I have moved the assertion message into the documentation to make it more clear.

src/DGtal/base/Alias.h
JacquesOlivierLachaud: Good idea ! I assume you wish to put = default ?

rolanddenis: Actually, I was thinking about deleting all those special members, since the copy operations are currently forbidden in the templates (see e.g. https://github.com/DGtal-team/DGtalScripts/blob/master/dev/models/TXXX.h)

It will give something like:
```
public:
XXX() = delete;
~XXX() = delete;
XXX( const XXX & ) = delete;
XXX( XXX && ) = delete;
XXX & operator= ( const XXX & ) = delete;
XXX & operator= ( XXX && ) = delete;
```
with appropriate documentation for each ones, and then letting the developer choosing between:
- keeping it deleted,
- switching it to the default behavior by replacing `delete` with `default`,
- implementing his own operations.

JacquesOlivierLachaud: Good for me.

rolanddenis: Ok, I will see what I can do :octocat:

dcoeurjo:
Fine for the doc as it is. merging.

PR #1148, opened 2016-03-16
dec cosmetic fix

elcerdo:
*Thanks a lot for contributing to DGtal, before submitting your PR, please fill up the description and make sure that all checkboxes are checked.*

# PR Description

Thore are just minor fixes

# Checklist

### DIFF
--- a/src/DGtal/io/Display3D.ih
+++ b/src/DGtal/io/Display3D.ih
@@ -808,10 +808,12 @@ DGtal::Display3D< Space ,KSpace >::addCone(const RealPoint &p1, const RealPoint 
   zab = p2[2]-p1[2];
 
   norm = sqrt( xab*xab + yab*yab + zab*zab);
-  assert (norm > 0);
-  xab = xab / norm;
-  yab = yab / norm;
-  zab = zab / norm;
+  if (norm > 0);
+	{
+			xab = xab / norm;
+			yab = yab / norm;
+			zab = zab / norm;
+	}
 
   //take a third point M different from A and B
   double xm = 0;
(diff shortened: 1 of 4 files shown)


elcerdo:
asserts trigger if drawing a null corner at origin...

src/DGtal/io/Display3D.ih
```diff
@@ -808,10 +808,12 @@ DGtal::Display3D< Space ,KSpace >::addCone(const RealPoint &p1, const RealPoint
   zab = p2[2]-p1[2];
 
   norm = sqrt( xab*xab + yab*yab + zab*zab);
-  assert (norm > 0);
```
dcoeurjo: change assert to ```FATAL_ERROR```instead ? what is the behavior if the norm is null?

src/DGtal/io/Display3D.ih
```diff
@@ -830,10 +832,12 @@ DGtal::Display3D< Space ,KSpace >::addCone(const RealPoint &p1, const RealPoint
   zam = zm-p1[2];
 
   norm = sqrt( xam*xam + yam*yam + zam*zam);
-  assert (norm > 0);
-  xam = xam / norm;
-  yam = yam / norm;
-  zam = zam / norm;
+  if (norm > 0)
```
dcoeurjo: same here.

src/DGtal/io/Display3D.ih
elcerdo: when norm=0, you are trying to draw an cone with height=0.
imo this should be ignored and certainly not trigger any assert or error.
my understanding of this assert is that it guards the subsequent division by zero from generating nan. this can be easily avoided by placing the division inside a 'if(norm>0' block.

dcoeurjo:
ok fine. merging. thanks

Issue #779, opened 2013-12-18
Bug when importing Image with ImageContainerBySTLMap and DicomReader


kerautret:
When applying:
typedef DGtal::ImageContainerBySTLMap<DGtal::Z3i::Domain, unsigned char> Image3D;
typedef DGtal::RescalingFunctor<int ,unsigned char > RescalFCT;

 Image3D imageVol = DicomReader< Image3D,  RescalFCT  >::importDicom(inputFilename, RescalFCT(-900,
                                                                                               530,
                                                                                               0, 255));

The resulting domain is totally wrong!... While replacing ImageContainerBySTLMap by ImageContainerBySTLVector it is ok !....

isivigno:
I just encountered the same kind of problem with the importPGM reader : 

typedef ImageContainerBySTLMap<Z2i::Domain, unsigned char> Image;
Image image = DGtal::PGMReader<Image>::importPGM("../circleR10modif.pgm");   
Board2D board;
typedef GrayscaleColorMap<unsigned char> Gray;
Display2DFactory::drawImage<Gray>(board, image, (unsigned int)0, (unsigned int)129);  
---> assertion fails 

The good news is that it works if you use the GenericReader: 
Image image = DGtal::GenericReader<Image>::import("../circleR10modif.pgm");

JacquesOlivierLachaud:
Should be fixed by #1151. Is it the case ?

kerautret:
I look it ;)

JacquesOlivierLachaud:
Thanks!

kerautret:
yes all looks fine !
With dicom2vol compiled with STLMap:

./converters/dicom2vol -i ../../DGtal/tests/samples/dicomSample/1629.dcm -o tmp.vol
Reading input dicom file ../../DGtal/tests/samples/dicomSample/1629.dcm [done] 
 converting into vol file...  [done] 
And the vol looks fine:
./visualisation/3dImageViewer -i tmp.vol  --thresholdImage -m 0 -M 63
Image loaded: [Image - STLVector] size=2130048 valuetype=1bytes Domain=[HyperRectDomain] = [[PointVector] {0, 0, 0}]x[[PointVector] {128, 127, 128}]

JacquesOlivierLachaud:
Perfect. Merging.

JacquesOlivierLachaud:
Closing issue

PR #1151, opened 2016-03-21
Fixing issue #779

rolanddenis:
# PR Description

The issue #779 comes from the fact that `ImageContainerBySTLMap` stores his domain by pointer.
But readers create domain on the stack before constructing images so that the pointer is invalid as soon as the reader ends. 

Therefore, a solution is to store the domain by value, like other image containers.
But, doing so (like in this PR), leads to compilation errors when the domain is not assignable. It is the case when using a `DigitalSetDomain`, like in `tests/geometry/volumes/distance/testPowerMap`.

In fact, this problem arises for all `CTrivialConstImage` models that store the domain by value, since it is a refinement of `boost::Assignable`.

Any suggestion ?

# Checklist

- [N/A] Doxygen documentation of the code completed (classes, methods, types, members...)

### DIFF
--- a/src/DGtal/base/CowPtr.h
+++ b/src/DGtal/base/CowPtr.h
@@ -90,6 +90,25 @@ public:
     const T* operator->()   const throw()   {return myPtr.get();}
     const T* get()          const throw()   {return myPtr.get();}
 
+    /* The following non-const methods are deactivated if T is a const type.
+     * The goal here is to avoid unecessary copies when it is known that
+     *   the T object will not be modified.
+     *
+     * The problem is that C++ uses the non-const methods whenever it's possible
+     *   (ie when CowPtr<T> is non-const), even if the full expression doesn't 
+     *   modify the object. A solution is to const_cast the CowPtr<T> before
+     *   using one of these methods to force the usage of the const versions above.
+     *
+     * However, in the trivial case when T is a const type, we can simplify this
+     *   by deactivating those methods.
+     *
+     * To do that, we use std::enable_if (see http://en.cppreference.com/w/cpp/types/enable_if )
+     *   in the template parameters (it is also possible in the return type or
+     *   as a parameter but it will change the signature). It depends on
+     *   the constness of T returned by the type trait std::is_const.
+     * The `typename U = T` is necessary in order that the SFINAE rule works at the overload
+     * resolution step.
+     */
     template < typename U = T, typename std::enable_if< ! std::is_const<U>::value >::type* = nullptr >
     T& operator*()                          {copy(); return *myPtr;}
 
--- a/src/DGtal/base/CowPtr.h
+++ b/src/DGtal/base/CowPtr.h
@@ -43,6 +43,7 @@
 //////////////////////////////////////////////////////////////////////////////
 // Inclusions
 #include <iostream>
+#include <type_traits>
 #include "DGtal/base/Common.h"
 #include "DGtal/base/CountedPtr.h"
 //////////////////////////////////////////////////////////////////////////////
@@ -89,8 +90,13 @@ public:
     const T* operator->()   const throw()   {return myPtr.get();}
     const T* get()          const throw()   {return myPtr.get();}
     
+    template < typename U = T, typename std::enable_if< ! std::is_const<U>::value >::type* = nullptr >
     T& operator*()                          {copy(); return *myPtr;}
+
+    template < typename U = T, typename std::enable_if< ! std::is_const<U>::value >::type* = nullptr >
     T* operator->()                         {copy(); return myPtr.get();}
+
+    template < typename U = T, typename std::enable_if< ! std::is_const<U>::value >::type* = nullptr >
     T* get()                                {copy(); return myPtr.get();}
 
     /**
--- a/src/DGtal/base/CowPtr.h
+++ b/src/DGtal/base/CowPtr.h
@@ -89,7 +89,7 @@ public:
     const T& operator*()    const throw()   {return *myPtr;}
     const T* operator->()   const throw()   {return myPtr.get();}
     const T* get()          const throw()   {return myPtr.get();}
-    
+
     template < typename U = T, typename std::enable_if< ! std::is_const<U>::value >::type* = nullptr >
     T& operator*()                          {copy(); return *myPtr;}
 
@@ -101,7 +101,7 @@ public:
 
     /**
        Equality operator ==
-       
+
        @param other any other pointer.
        @return 'true' if pointed address is equal to \a other.
     */
@@ -112,7 +112,7 @@ public:
 
     /**
        Inequality operator !=
-       
+
        @param other any other pointer.
        @return 'true' if pointed address is different from \a other.
     */
--- a/src/DGtal/images/ImageContainerBySTLMap.h
+++ b/src/DGtal/images/ImageContainerBySTLMap.h
@@ -112,6 +112,10 @@ namespace DGtal
     typedef typename Domain::Dimension Dimension;
     typedef Point Vertex;
 
+    // Pointer to the (const) Domain given at construction.
+    typedef const Domain * DomainPtr;
+
+
     /// static constants
     static const typename Domain::Dimension dimension;
 
@@ -127,8 +131,10 @@ namespace DGtal
     /////////////////// Data members //////////////////
   private:
 
-    /// Image domain
-    Domain myDomain;
+    /// Counted pointer on the image domain,
+    /// Since the domain is not mutable, not assignable,
+    /// it is shared by all the copies of *this
+    DomainPtr myDomainPtr;
 
     /// Default value
     Value myDefaultValue;
@@ -144,7 +150,7 @@ namespace DGtal
      * @param aValue a default value associated to the domain points
      * that are not contained in the underlying map.
      */
-    ImageContainerBySTLMap( const Domain& aDomain, const Value& aValue = 0);
+    ImageContainerBySTLMap( ConstAlias<Domain> aDomain, const Value& aValue = 0);
 
     /**
      * Copy operator
--- a/src/DGtal/images/ImageContainerBySTLMap.h
+++ b/src/DGtal/images/ImageContainerBySTLMap.h
@@ -112,10 +112,6 @@ namespace DGtal
     typedef typename Domain::Dimension Dimension;
     typedef Point Vertex;
 
-    // Pointer to the (const) Domain given at construction.
-    typedef const Domain * DomainPtr;
-
-
     /// static constants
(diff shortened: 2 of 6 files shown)


phcerdan:
Not really aware of all the details here, but what about using std::shared_ptr instead of raw pointers?

Passing by shared_ptr only when lifetime, or ownership is involved, and by raw pointer std::shared_ptr::get()  to interact with non-owning functions/classes. [cpp core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r30-take-smart-pointers-as-parameters-only-to-explicitly-express-lifetime-semantics)

From [boost::shared_ptr](http://www.boost.org/doc/libs/1_57_0/libs/smart_ptr/shared_ptr.htm) it meets the CopyConstructible, MoveConstructible, CopyAssignable and MoveAssignable requirements.

dcoeurjo:
We do have smart pointers in DGtal (http://dgtal.org/doc/nightly/moduleCloneAndReference.html). This issue seems to be related to bad object lifetime  management in this container. 
I'll have a look but that's not a good idea to copy domain by value since they can be "heavy".

rolanddenis:
I agree that `std::shared_ptr` (or similar) is the right choice if we choose to keep pointer to the domain.
The problem is that, at this time, not all the image containers use the same way to refer to the domain and thus, it is difficult to write a generic image importer. 
So there is some solutions:
- every image container stores domain by (smart) pointer, but that need a lot of modifications in the code;
- every image container stores domain by value and we add a kind of domain proxy class that handles an underlying domain by (smart) pointer, for the cases when domains are heavy;
- a mixed solution where images that store his domain by pointer, feature 2 constructors : one that copy a domain into a new `shared_ptr` (to ensure compatibility with current code), and one that take directly a `shared_ptr`.

I suppose that we choose the third solution ? Perhaps with `CountedPtr` ?

src/DGtal/images/ImageContainerBySTLMap.h
```diff
@@ -78,102 +79,109 @@ namespace DGtal
...
     /////////////////// Data members //////////////////
-  private: 
+  private:
 
-    /// Counted pointer on the image domain,
+    /// Shared pointer on the image domain,
     /// Since the domain is not mutable, not assignable,
     /// it is shared by all the copies of *this
-    DomainPtr myDomainPtr;
+    std::shared_ptr<const Domain> myDomainPtr;
```
dcoeurjo: Couldn't it be a DGtal::CountedPtr ? 
just for consistency..

src/DGtal/images/ImageContainerBySTLMap.h
```diff
@@ -78,102 +79,109 @@ namespace DGtal
...
+     * Constructor from a Domain.
+     *
+     * If Domain is a heavy type, prefer giving a smart pointer on the domain.
+     *
      * @param aDomain the image domain.
      * @param aValue a default value associated to the domain points
      * that are not contained in the underlying map.
      */
-    ImageContainerBySTLMap( ConstAlias<Domain> aDomain, const Value& aValue = 0);
+    ImageContainerBySTLMap( const Domain & aDomain, const Value& aValue = 0);
```
dcoeurjo: I think I do prefer ConstAlias version: with "const &" the interface does not tell the user if the content is copied or aliased.
That was the starting point of the Alias/ConstAlias creation.

dcoeurjo: well you describe this in the doc "const&" vs. "pointer" constructor.

dcoeurjo:
Sorry for being late on this PR. Let me start with some general comments:

* all other image containers *should* have their constructor in the form

````
ImageContainer(ConstAlias<Domain> aDomain);
````
and member class could, for instance (several other options are possible as members, see doc page), be:
  
```` 
const Domain * myDomain;
````
(most image containers have been created before Const/Alias stuff). Lifetime of the domain must be larger than the container lifetime (i.e.: no domains on the stack)

* Thanks to some previous PR, @rolanddenis , if the user does a ```Image image(Domain(a,b))``` an warning should be raised, am I right ? 
* Plenty of smart pointers are already available in DGtal, I would recommend to use them
* Heavy/light data management int DGtal is something we have discussed a lot and maybe the solution we have right now is not satisfactory. 
* Would you suggest to have both 
 * a constructor with copy of the domain
 * a constructor with an (const) alias of the domain
for all containers?

I know that @elcerdo has very different point of view of the data management: constructors always copy the data and when the user don't want to have memory copies, he/she wraps the data into a lightweight facade (with exact same interface)... It is not was is currently implemented in DGtal but that's maybe part of the discussion.

(ping @JacquesOlivierLachaud )

File: src/DGtal/images/ImageContainerBySTLMap.h (hunk 2 above)
Comment (rolanddenis) [inline src/DGtal/images/ImageContainerBySTLMap.h]: Using here `ConstAlias` would disallow inline construction of the domain (it is sometimes the case in DGtal). I could also pass it by value but since domains are not moveable, this will imply an additional copy...

I can also make it more clear in the documentation that the domain is copied and not aliased.

File: src/DGtal/images/ImageContainerBySTLMap.h (hunk 1 above)
Comment (rolanddenis) [inline src/DGtal/images/ImageContainerBySTLMap.h]: Yes it can ! However  `CountedPtr` has no conversion from `CountedPtr<T>` to `CountedPtr<const T>` that may be necessary here. I will think about adding it...

File: src/DGtal/images/ImageContainerBySTLMap.h (hunk 2 above)
Comment (dcoeurjo) [inline src/DGtal/images/ImageContainerBySTLMap.h]: I agree but that is the case for many classes.. (e.g. CSG construction of shapes). And I agree with you that it is painful for the user.

Comment (rolanddenis) [inline src/DGtal/images/ImageContainerBySTLMap.h]: Ok, it is not a right argument :octocat: 

I try again: using `ConstAlias` here says to the user that his domain is aliased in the image and thus, he may think that it is a good choice for heavy domains (nobody reads the doc :trollface:).

**Comment by rolanddenis** [comment]
> all other image containers should have their constructor in the form
```ImageContainer(ConstAlias<Domain> aDomain);```

OK but that means that the caller controls the lifetime of `aDomain` but it is not possible for, e.g., image readers that should otherwise returns the image and a smart pointer on the domain.

> Thanks to some previous PR, @rolanddenis , if the user does a `Image image(Domain(a,b))` an warning should be raised, am I right ?

No because `ImageContainerBySTLMap` is an exception among image containers: many of them (`ImageContainerBySTLVector`, `ImageContainerByHashTree` and mayby `ImageContainerByITKImage` ?) only handle `HyperRectDomain` domain type (that is not a heavy type) and store it by value.

> Would you suggest to have both
    - a constructor with copy of the domain
    - a constructor with an (const) alias of the domain for all containers?

I don't think this is a good solution but I have choose it for compatibility reasons with current code. I don't like the idea that classes have to deal with their parameters size when they do not control their type (because of a template) and thus, I like @elcerdo solution even if it requires more works for the user.

A mixed solution may be what some linear algebra library uses (like `Blitz++` if I'm not wrong): copy construction, like `Array A(B)`, aliases data (`A` and `B` refers to the same array), while copy assignment really copies the data. But it is probably not what the user excepts from a copy constructor and I am not sure that this have no side effects...

**Comment by JacquesOlivierLachaud** [comment]
Just a silly C++11 question... The rvalue constructor of ConstAlias prevents the passing of temp objects.
And an image reader can return a CountedPtr<Domain> and the ConstAlias should work fine.
Have I missed something ?

**Comment by rolanddenis** [comment]
After discussing with @JacquesOlivierLachaud, it seems that using `Clone` together with `CowPtr` is an appropriate choice in our case and that avoids using two different constructors.
In addition, I have modified `CowPtr<T>` to deactivate non-const methods when `T` is a constant type, in order to avoid unnecessary copies (NB: `C++` calls non-const methods whenever it's possible, even if the expression does not modify the object. So, it would trigger a copy of the owned `T` if the instance of `CowPtr<T>` is non-const, even if `T` is constant.)

File: src/DGtal/base/CowPtr.h [hunk 3]
```diff
@@ -88,14 +89,19 @@ namespace DGtal
     const T& operator*()    const throw()   {return *myPtr;}
     const T* operator->()   const throw()   {return myPtr.get();}
     const T* get()          const throw()   {return myPtr.get();}
-    
+
+    template < typename U = T, typename std::enable_if< ! std::is_const<U>::value >::type* = nullptr >
```
JacquesOlivierLachaud: Do you mind commenting a bit these lines ? (Like forces the compiler to choose the const version without copy whenever possible)

tests/geometry/volumes/distance/testReducedMedialAxis.cpp
```diff
@@ -52,26 +53,26 @@ bool testReducedMedialAxis()
...
   set.insertNew(Z2i::Point(7,7));
-  DigitalSetDomain<DigitalSetBySTLSet<Z2i::Domain > > setDomain(set); 
-  
-  typedef ImageContainerBySTLMap<DigitalSetDomain<DigitalSetBySTLSet<Z2i::Domain > > , DGtal::int64_t> Image;
-  Image image(setDomain);
-  
+
+  using SetDomain = DigitalSetDomain< DigitalSetBySTLSet<Z2i::Domain > >;
+  using Image = ImageContainerBySTLMap< SetDomain , DGtal::int64_t>;
+  Image image( CountedPtr<const SetDomain>( new SetDomain( set ) ) );
```
JacquesOlivierLachaud: If you wish here, you can just pass "new SetDomain( set )" which will be acquired by the image because of the Clone<Domain>.

src/DGtal/base/CowPtr.h
rolanddenis: I think it was a very comprehensible line, especially the `typename U = T`:trollface: 
Ok, I will add a comment :wink: ( in doxygen style or just for the devs ? )

tests/geometry/volumes/distance/testReducedMedialAxis.cpp
rolanddenis: Right !
I was also wondering that we can simply write:
`Image image( SetDomain( set ) )`
that will move the temporary up to the `CowPtr`. The problem is that `DigitalSetBySTLSet`is not yet moveable...

src/DGtal/base/CowPtr.h
JacquesOlivierLachaud: Just for devs I think.

tests/geometry/volumes/distance/testReducedMedialAxis.cpp
JacquesOlivierLachaud: Should work but if the class is not moveable it will copy it, no ?

PR #1153, opened 2016-03-25
fix compile in itk reader/writer

kerautret:
# PR Description

fix compile in itk reader/writer

# Checklist

- [n.a] Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).
- [n.a] Doxygen documentation of the code completed (classes, methods, types, members...)
- [n.a] Documentation module page added or updated.

### DIFF
--- a/src/DGtal/io/readers/ITKReader.ih
+++ b/src/DGtal/io/readers/ITKReader.ih
@@ -64,7 +64,8 @@ namespace DGtal {
     const Domain& domain = dgtal_itk_image.domain();
 
     typedef ConstImageAdapter<DGtalITKImage, Domain, functors::Identity, Value, Functor> AdaptedImage;
-    const AdaptedImage adapted(dgtal_itk_image, domain, functors::Identity(), aFunctor);
+    const functors::Identity identityFunctor{};
+    const AdaptedImage adapted(dgtal_itk_image, domain, identityFunctor, aFunctor);
 
     Image image(domain);
     std::copy(adapted.constRange().begin(), adapted.constRange().end(), image.range().outputIterator());
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,8 @@
    constructor, like from a rvalue reference. Adding ConstAlias in many classes
    that need it.
    (Roland Denis, [#1140](https://github.com/DGtal-team/DGtal/pull/1140))
+   (With ITK related compilation fix, Bertrand Kerautret
+   [#1153](https://github.com/DGtal-team/DGtal/pull/1153))
 
 - *IO Package*
  - Add the possibility to interact in QGLViewer Viewer3D class with the voxel
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -69,6 +69,12 @@
     was wrong on 3d and on closed Khalimsky space.
     (Bertrand Kerautret, [#1156](https://github.com/DGtal-team/DGtal/pull/1156))
 
+- *Shape Package*
+  - Fix a tubular mesh construction problem (missing faces) which appears
+    when the center line is oriented in a main axis direction (in
+    createTubularMesh()). Also improves and fixes the face construction.
+    (Bertrand Kerautret, [#1157](https://github.com/DGtal-team/DGtal/pull/1157))
+
 # DGtal 0.9.1
 
 ## New Features / Critical Changes
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -65,7 +65,9 @@
   - Fix wrong starting point for surface tracking in example code
     frontierAndBoundary.cpp.
     (Roland Denis, [#1144](https://github.com/DGtal-team/DGtal/pull/1144))
-
+  - Fix interior/exterior fill methods of topology/helpers/Surfaces class which
+    was wrong on 3d and on closed Khalimsky space.
+    (Bertrand Kerautret, [#1156](https://github.com/DGtal-team/DGtal/pull/1156))
 
 # DGtal 0.9.1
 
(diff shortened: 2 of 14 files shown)


src/DGtal/io/readers/ITKReader.ih
```diff
@@ -64,7 +64,8 @@ namespace DGtal {
     const Domain& domain = dgtal_itk_image.domain();
 
     typedef ConstImageAdapter<DGtalITKImage, Domain, functors::Identity, Value, Functor> AdaptedImage;
-    const AdaptedImage adapted(dgtal_itk_image, domain, functors::Identity(), aFunctor);
+    const functors::Identity identityFunctor{};
```
dcoeurjo: why ```{}```?

dcoeurjo:
beside my minor comment, this PR is fine.
Thanks

src/DGtal/io/readers/ITKReader.ih
kerautret: I apply the same change of @rolanddenis ;) looks fine without with my compilo but I suppose that there a reason.

rolanddenis: :smile: 
In that case, it is exactly the same as without `{}`: it calls the constructor without argument. It is a new syntax from `C++11`. It enhances the construction of object and, especially, it allows me to replace a common mistake that I make:
```const functors::Identity identityFunctor();```
by a valid syntax:
```const functors::Identity identityFunctor{};```
:octocat:

Comment (dcoeurjo) [inline src/DGtal/io/readers/ITKReader.ih]: Well, I would have preferred to get the () removed instead of something that can be ambiguous for non-c11 dev. 
(Not sure to see why this new syntax enhance the construction.. I'll have to read the doc;))

**Comment by kerautret** [comment]
ok @dcoeurjo I have remove the {} since @rolanddenis confirms that in this case it is equivalent.
Perhaps we can merge and change other {} in other PR? (there also some in DGTalTools)

**Comment by dcoeurjo** [comment]
Yes I agree. Thanks

**Comment by dcoeurjo** [comment]
(Just waiting for Travis ;))

**Comment by rolanddenis** [comment]
Don't merge yet: do you have try to compile it with clang ? It should fail because of the missing `{}`. 
I can't try this specific file (my `ITK` is not compiled with `C++11`) but this simple example:
```
#include "DGtal/base/BasicFunctors.h"
int main()
{
  const DGtal::functors::Identity id;
  return 0;
}
```
failed with `clang++` and it is not a bug.
(explanation coming)

rolanddenis:
There is a small rule when using default initialization with constant object (§8.5/7):
> If a program calls for the default initialization of an object of a const-qualified type `T`, `T` shall be a class type with a user-provided default constructor.

It seems that this rule exists to avoid invalid initialization of POD class with `const` qualifier.
In order to compile, there is many solutions:
- declaring a user-provided default constructor. "user-provided" means that the constructor is not set to the default (with `= default` at the end of the first declaration). Therefore, the class becomes non-POD.
- removing the `const` qualifier (against the `const` correctness).
- explicitly using the "value initialization" (instead of the "default initialization") by adding `{}`:
`const DGtal::functors::Identity id{};`
or, before `C++11`:
`const DGtal::functors::Identity id = DGtal::functors::Identity();`

`gcc`does not complain because it implements a proposed change in the draft, coming after the `C++11`standard that weaken this rule for class without members (see http://stackoverflow.com/questions/26077807/why-does-gcc-allow-a-const-object-without-a-user-declared-default-constructor-bu and http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#253 ). This change seems to be still not applied in the last draft.

Refs:
- http://stackoverflow.com/questions/21900237/do-i-really-need-to-implement-user-provided-constructor-for-const-objects
- http://stackoverflow.com/questions/7411515/why-does-c-require-a-user-provided-default-constructor-to-default-construct-a
- http://en.cppreference.com/w/cpp/language/default_initialization
- http://en.cppreference.com/w/cpp/language/value_initialization
- https://akrzemi1.wordpress.com/2013/09/10/value-initialization-with-c/

kerautret:
thanks @rolanddenis , yes you are right it fails, perhaps I wrongly switch branch by testing by removing it, sorry ;( !
So I revert to the initial good version ;)

dcoeurjo:
Thanks for the details! I've learned a lot of things. 
So what should we do? Should we recommend to add default constructors to our functors or  let the user deal with the {} (or its pre-c++11 version) ?

kerautret:
perhaps the default constructor is better but we will lost the POD of  Identity, but I don't know if it is a problem to be no more Plain Old Data ?

rolanddenis:
I found that strange to provide a default constructor that do nothing in a class that owns nothing (and that could have only `static` methods). If using `{}` is really a problem, I prefer removing the `const`. A maybe better solution is to mix the old fashion way with `auto`:
`const auto id = DGtal::functors::Identity();`
?

kerautret:
Perhaps the {} is fine from the user point of view, since the message of Clang is really explicit:
capture d ecran 2016-03-31 a 22 58 12
and also the auto should be fine..

dcoeurjo:
Ok, you're both right. Let's keep the {} for const instanciations. 
Thanks for the discussion :smile:

So, can i merge this branch or you need to revert your edits, Bertrand ?

kerautret:
I revert it just now, I was waiting the end of discussion ;)

dcoeurjo:
:ok:

kerautret:
done

kerautret:
@dcoeurjo can we merge?

dcoeurjo:
Yep merging. Thanks

kerautret:
thanks! ;)

dcoeurjo:
ok.. there is an issue.. could you please merge the master back ?

kerautret:
it looks like it was just a meriging bug ! ;)

PR #1147, opened 2016-03-16
VoxelComplex: CriticalKernels framework

phcerdan:
# PR Description

Implementation of Critical Kernel framework based on M.Couprie and G.Bertrand articles:
- [Asymmetric parallel 3D thinning scheme and algorithms based on isthmuses.](https://hal.archives-ouvertes.fr/hal-01217974/document)
- [Powerful Parallel and Symmetric 3D Thinning Schemes Based on Critical Kernels](https://hal.archives-ouvertes.fr/hal-00828450/)
- [A 3D Sequential Thinning Scheme Based on Critical Kernels](https://hal.archives-ouvertes.fr/hal-01217952/file/Critical3Dseq.pdf)
- [x] Add general description of framework for docs.
- [x] Add LookUpTable interface from #1155  (speeds an order of magnitude)

[CLI script for testing in any image (future PR in DGtalTools)](https://github.com/phcerdan/dgtal_criticalkernels)
## Implementation Details
- VoxelComplex inherits from CubicalComplex, but no override or hide any method.
- It has an Object member storing the voxels(spels) of an input image. This allows the use of the isSimple method from Object, and also the ObjectBoostGraphInterface.
- CriticalCliques:
  A clique is a set of Voxels.
  K_0, K_1, K_2, and K_3 return a pair<bool, Clique> where Clique=CubicalComplex, and the bool is true if the Clique is critical (as described in papers).
  
  K_3 applies to Voxels, and it uses the isSimple method of Object.
  
  K_2 applies to surfels, K_1 to linels, and K_0 to pointels.

In VoxelComplexFunctions a couple of thinning methods are implemented in a functional style, taking a VoxelComplex, a Select function, and a Skel function.

**Select function:** Select a voxel from a clique (set of voxels)
- Random.
- First: just the first voxel of the set.
- MaxValue: Wrap to use with a lambda, to use a DGtal::DistanceTransformation to select the voxel with greatest distance value.

**Skel function:** Predicate to check if input voxel is part of the skeleton that we are interested to preserve.
- Ultimate: Null, preserve only to keep same topology. The result of the thinning in a connected object is just one voxel.
- End: preserve if voxel has only one neighbor.
- OneIsthmus: preserve if voxel is oneIsthmus, as described in articles.
- TwoIsthmus: idem, this conserves planar structures.
- IsIsthmus : (OneIsthmus || TwoIsthmus)
[...]

PR #1155, opened 2016-03-27
Add LookUpTableFunctions

phcerdan:
# Add LookUpTableFunctions

New functions to deal with pre-computed look up tables.
The motivation is to speed up large computation that use predicate
functions depending on voxel neighborhood, such as isSimple.

#### Add new folder /topology/tables containing:
* NeighborhoodConfigurationGenerators, function used to generate the tables.
* compressed .tar.gz tables for simplicity.

#### Modify example generateSimplicityTables3d:
* using functions from new header.
* create tables for different topologies.
* user can choose 26_6,6_26, 18_6 and 6_18 topologies.
* Generated tables are heavy, 64MB, so they have been added
 to the source code after compressed using:
    `cmake -E tar cvzf table26_6.tar.gz simple_table26_6.txt` resulting in a light (<500KB) tar file.

#### Add CMake rules:
* TablesConfig.cmake dealing with decompress,
    copy and install tables. Linked in root CMakeLists.txt
* ConfigTables.h.in . CONFIGURE_FILE with strings pointing
to all available tables. The generated ConfigTables.h
change location of tables depending on Build or Install.

# Checklist


JacquesOlivierLachaud:
It seems a good idea to me.
I was wondering if we should not do the same for the 2D case ;-)

phcerdan:
The only concern I have is that every 3D table (with 2^26 configurations) is 64 MB after decompression.
Not sure if it is better to decompress at build/install time (this PR) or include an external library, such as zlib or minizip, to decompress at run time. (there is a [boost filter that works as a zlib wrap](http://www.boost.org/doc/libs/1_50_0/libs/iostreams/doc/classes/zlib.html) )

I don't know how keen you are to add the zlib external dependency. For a few tables this approach looks all right though.

@JacquesOlivierLachaud Yeah, 2D tables are straight forward to add, and much lighter. I will push them.

src/DGtal/topology/tables/LookUpTableFunctions.h
```diff
@@ -0,0 +1,188 @@
...
+#include <bitset>
+#include <memory>
+#include "DGtal/topology/Object.h"
+#include "DGtal/helpers/StdDefs.h"
+#include <unordered_map>
+#include "boost/dynamic_bitset.hpp"
+
+namespace DGtal
+{
+  namespace lut {
```
dcoeurjo: global functions should be in the ```DGtal::functions::``` namespace.

dcoeurjo: ```DGtal::functions::simplicity```has namespace ?  ```lut``` is too generic IMHO

src/DGtal/topology/tables/LookUpTableFunctions.ih
```diff
@@ -0,0 +1,200 @@
...
+    TObject shape( dt, shapeSet );
+    unsigned int k = 0;
+    for ( DomainConstIterator it = domain.begin(); it != domain.end(); ++it )
+      if ( *it != c ) ++k;
+    ASSERT( ( k < 32 )
+        && "[generateSimplicityTable] number of configurations is too high." );
+    unsigned int nbCfg = 1 << k;
+    for ( unsigned int cfg = 0; cfg < nbCfg; ++cfg )
+    {
+      if ( ( cfg % 1000 ) == 0 )
```
dcoeurjo: maybe we should put these traces between ```#ifdef VERBOSE```  (http://dgtal.org/doc/nightly/moduleBuildDGtal.html#cmakeDGtal)

tests/topology/testLookUpTableFunctions.cpp
```diff
@@ -0,0 +1,298 @@
...
+  }
+  diamond_set.erase( c );
+  KSpace K;
+  K.init( domain.lowerBound(),
+      domain.upperBound(), true);
+
+  return TObject(dt, diamond_set);
+}
+
+TEST_CASE("MapPointToBitMask 3D", "[map][mask][3D]" )
```
dcoeurjo: could you be a bit more descriptive in the test case definition text ?

src/DGtal/topology/tables/LookUpTableFunctions.h
dcoeurjo: In a similar way, the filename ```LookUpTableFunctions``` is not great.. I have the feeling that it could be confusing. What do you think?

src/DGtal/topology/tables/LookUpTableFunctions.ih
phcerdan: generateSimplicityTable is only used in examples `examples/topology/generateSimplicityTables2D.cpp` and `...Tables3D.cpp` to generate the files. I have found the progress bar useful, because the process can be slow in 3D. But of course, whatever you prefer, I didn't know you had verbosity flag at compile time. Your call!

src/DGtal/topology/tables/LookUpTableFunctions.h
phcerdan: Good point, I will add the functions:: namespace asap.

The motivation of this PR is related to the other PR #1147 , where we use look up tables to calculate the (way more expensive) isIsthmus that depends on the neighborhood of voxels. If `lut` is too generic, `simplicity` is not able to capture the `isthmus` case. Maybe `tables::`?

I am completely open to suggestions of changing the name, but I cannot really think in anything better or more clarifying/descriptive. The user probably will only interact with this using the `loadTable("string_variable_pointing_to_table_file")` from `ConfigTable.h.in`. Other names: `NeigborhoodConfigurations`, `PreComputedTables`, `PreComputedConfiguration` ? Not sure, sorry!!

tests/topology/testLookUpTableFunctions.cpp
phcerdan: Thanks, I pushed descriptive messages for all tests.

src/DGtal/topology/tables/LookUpTableFunctions.h
dcoeurjo: :smile: no idea, too

```NeigborhoodConfigurations```  is ok for the filename but I don't know for the namespace... ```neighborhood::``` ?

File: src/DGtal/topology/tables/LookUpTableFunctions.ih (hunk 2 above)
Comment (dcoeurjo) [inline src/DGtal/topology/tables/LookUpTableFunctions.ih]: well.. I agree, this function is not supposed to be used a lot.. let's keep the trace as is.

**Comment by dcoeurjo** [comment]
could you please also consider to add something in the doc (```moduleDigitalTopology.dox```, last sections) ?
thx

**Comment by dcoeurjo** [comment]
BTW, maybe the dgtaltools should also be updated (ping @kerautret )

File: src/DGtal/topology/tables/LookUpTableFunctions.h (hunk 1 above)
Comment (JacquesOlivierLachaud) [inline src/DGtal/topology/tables/LookUpTableFunctions.h]: Why not a name related to the objective of lut... here topology, homotopy, simplicity, etc ... as you wish 

----- Mail original -----

De: "David Coeurjolly" <notifications@github.com> 
À: "DGtal-team/DGtal" <DGtal@noreply.github.com> 
Cc: "Jacques-Olivier Lachaud" <jacques-olivier.lachaud@univ-savoie.fr> 
Envoyé: Lundi 11 Avril 2016 11:16:01 
Objet: Re: [DGtal-team/DGtal] Add LookUpTableFunctions (#1155) 



In src/DGtal/topology/tables/LookUpTableFunctions.h : 
> +#if !defined LookUpTableFunctions_h
> +/** Prevents repeated inclusion of headers. */
> +#define LookUpTableFunctions_h
> +//////////////////////////////////////////////////////////////////////////////
> +// Inclusions
> +#include <iostream>
> +#include <bitset>
> +#include <memory>
> +#include "DGtal/topology/Object.h"
> +#include "DGtal/helpers/StdDefs.h"
> +#include <unordered_map>
> +#include "boost/dynamic_bitset.hpp"
> +
> +namespace DGtal
> +{
> +  namespace lut { 


no idea, too 

NeigborhoodConfigurations is ok for the filename but I don't know for the namespace... neighborhood:: ?

**Comment by phcerdan** [comment]
What do you prefer `NeigborhoodConfigurations` or `NeighborhoodPredicates` for the file name?
I am moving `NeigborhoodXXX.h/ih`  into the folder `topology/` and put the compressed tables and the configure file `NeighborhoodTables.h.in` in `topology/tables` ... better?

I like the split of the namespaces for the different tables.
`simplicity` sounds great, but for the Isthmus case? `topology` ? Or `isthmuscity` :)?

I will update the docs, and the log meanwhile.

**Comment by phcerdan** [comment]
Last two commits:
Rename files.
Add object interface to tables with setTable method: it sets CowPtr<T> members with a Clone<T> argument.
Modified isSimple to use the tables if setTable has been called.
No new constructors added to Object.
Add docs with an extract of testObject.cpp using the setTable interface.
Split header into helpers such as getOccupancyConfiguration for object and complex (to avoid recursive inclusion). And also the generateTables is there.

Maybe I can change the output of loadTable from std::shared_ptr<dynamic_bitset> to CountedPtr<dynamic_bitset> to fit better with the Clone<dynamic_bitset> argument of setTable.

File: cmake/NeighborhoodTablesConfig.cmake [hunk 4]
```diff
@@ -0,0 +1,37 @@
...
+
+#--- Unpack files at cmake configuration time.
+file(GLOB DGTAL_TABLES_COMPRESSED
+     ${PROJECT_SOURCE_DIR}/src/DGtal/topology/tables/*.tar.gz)
+set(unzip_folder ${PROJECT_BINARY_DIR}/src/DGtal/topology/tables/)
+message(STATUS "Decompressing look up tables to: ${unzip_folder}")
+foreach(zip_table ${DGTAL_TABLES_COMPRESSED})
+  execute_process(
+    COMMAND           ${CMAKE_COMMAND} -E tar xzf ${zip_table}
+    WORKING_DIRECTORY ${unzip_folder}
```
JacquesOlivierLachaud: Does this work on Windows ? Should we add something in the Windows install guide (like install tar archive tool) ?

examples/topology/generateSimplicityTables3D.cpp
```diff
@@ -30,96 +30,52 @@
...
+  }
   trace.endBlock();
 
-  ofstream file26_6( "simplicity_table6_26.txt" );
-  file26_6 << *table26_6;
-  file26_6.close();
+  string filename = "simplicity_table" + input_str + ".txt";
+  ofstream file( filename  );
+  file << *table;
+  file.close();
```
JacquesOlivierLachaud: By the way, did you check on some real examples that my function Object::isSimple is correct for all topology pairs ? I had problem at some point (which I tried to correct at best) but I never completely check these methods. Well, they were able to thin homotopically reasonnable size examples, and for 6_18 and 6_26, they produce thick skeleton, while being thinner for other adjacencies.

src/DGtal/topology/NeighborhoodConfigurations.h
```diff
@@ -0,0 +1,118 @@
...
+   * @return smart ptr of map[neighbor_configuration] -> bool
+   *
+   * @note The tables were calculated using the generateTableXXX examples.
+   * Compressed files of the tables are distributed in the source code.
+   * At build or install time, the header
+   * "DGtal/topology/tables/ConfigTables.h" is generated.
+   * It has const strings variables with the file names of the tables.
+   */
+  inline
+  DGtal::CountedPtr< boost::dynamic_bitset<> >
```
JacquesOlivierLachaud: Cool, there is a dynamic_bitset now in C++. I had to write one in 2002, and the "not operation" was kind of a pain because of trailing bits.

src/DGtal/topology/NeighborhoodConfigurations.h
```diff
@@ -0,0 +1,118 @@
...
+   *
+   * @tparam TPoint type of point to create map and input the desired dimension.
+   *
+   * @return map[Point]->configuration smart pointer.
+   */
+  template<typename TPoint>
+  inline
+  DGtal::CountedPtr<
+  std::unordered_map<TPoint, unsigned int > >
+  mapPointToBitMask();
```
JacquesOlivierLachaud: At first read, I do not understand this function. How can one point provides a whole configuration (either 3x3 or 3x3x3 points) ?

JacquesOlivierLachaud: After reading the implementation, I got what it does. I would not say that it maps points in a neighborhood to a configuration. I would say that it maps any point in the neighborhood of (0,0,0) to its corresponding mask in the configuration table. What do you think ?

cmake/NeighborhoodTablesConfig.cmake
phcerdan: [CMake](https://cmake.org/cmake/help/v3.2/manual/cmake.1.html) provides it, and it is mulit-platform, but I am not really sure what version of cmake is required.

src/DGtal/topology/NeighborhoodConfigurations.h
phcerdan: I think that is a really good description, which goes directly to the docs of the function. What name do you suggest for the function? This is really a "helper" for the look up table, but it is used every time you implement something with a table, so a good name will definitely help.

src/DGtal/topology/NeighborhoodConfigurations.h
phcerdan: Dealing with bits is quite painful... `boost::dynamic_bitset` is handy, also I have read complains about the `bool` specialization of std::vector.

examples/topology/generateSimplicityTables3D.cpp
phcerdan: No, I haven't checked it, I have only used 26_6 so far.

src/DGtal/topology/NeighborhoodConfigurations.h
JacquesOlivierLachaud: I agree with you, this is a very generic function. However, if we wish to have a generic approach to look-up tables, then we should have a generic class LookUpTable, mapping configurations to values (possibly bool), having load and save, etc... It is a bit more work, no ?

src/DGtal/topology/NeighborhoodConfigurationsHelpers.h
```diff
@@ -0,0 +1,132 @@
...
+   * @param input_obj input object. Used to check what points are occupied.
+   * @param center of the neighborhood. It doesn't matter if center belongs or not to \b input_object.
+   * @param mapPointToMask map[Point]->configuration, where Point is inside a 3x3 cube centered in {0,0,0} in 3D, or 2x2 in 2D.
+   * @see mapPointToBitMask
+   *
+   * @return bit configuration
+   */
+  template<typename TObject>
+  unsigned int
+  getOccupancyConfiguration(
```
JacquesOlivierLachaud: Object could be replaced by any model of CPointPredicate, no ?

src/DGtal/topology/NeighborhoodConfigurationsHelpers.h
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Get the occupancy configuration of the neighborhood of a point in a cubical
+   * complex.
+   * The neighborhood size is considered DxD for dimension D of the point
+   * (ie 3x3 cube for 3D point).
+   *
+   * @tparam TComplex Complex type.
+
+   * @param input_complex input complex. Used to check what points are occupied.
```
JacquesOlivierLachaud: Is it points or cells here ?

JacquesOlivierLachaud: Ok, I read the implementation. This method is valid for spels solely. Perhaps the method could be renamed as getSpelOccupancyConfiguration ?

src/DGtal/topology/Object.ih
```diff
@@ -60,7 +61,12 @@
 template <typename TDigitalTopology, typename TDigitalSet>
 inline
 DGtal::Object<TDigitalTopology, TDigitalSet>::Object()
-  : myTopo( 0 ), myPointSet( 0 ), myConnectedness( UNKNOWN )
+  : myTopo( nullptr ),
```
JacquesOlivierLachaud: I understand that nullptr is the norm for 0 since C++11, while 0 for the norm for NULL before. I am really looking for more stability in C++ !

src/DGtal/topology/Object.ih
```diff
@@ -891,6 +920,28 @@ DGtal::Object<TDigitalTopology, TDigitalSet>
...
+::isSimpleFromTable( const Point & center, const boost::dynamic_bitset<> & input_table, const std::unordered_map<Point,unsigned int> & pointMap ) const
+{
+  // Get Neighborhood Occupancy Configuration
+    using DomainConstIterator = typename Domain::ConstIterator;
+    Point p1 = Point::diagonal( -1 );
+    Point p2 = Point::diagonal(  1 );
+    Point c = Point::diagonal( 0 );
+    Domain ddomain( p1, p2 );
+    const auto & not_found( pointSet().end() );
+    unsigned int cfg{0};
```
JacquesOlivierLachaud: Perhaps you should define a type NeighborhoodConfiguration (in helpers/Neigh...) that is unsigned int and then use it where appropriate. I agree that it will be unsigned int32_t in all practical cases.

phcerdan: I tried, but I get a circular dependency because header of getOccupancyConfiguration depends on Object.h.

JacquesOlivierLachaud: You are right. But in fact getOccupancyConfiguration does not really need Object (only a CPointPredicate).
Only generateSimplicityCfg requires Object. Anyway, it is not important.

tests/topology/testObject.cpp
```diff
@@ -738,6 +740,82 @@ bool testObjectGraph()
...
+  // Greedy thinning.
+  DGtal::uint64_t nb_simple;
+  trace.beginBlock ( "Greedy homotopic thinning with table..." );
+  int layer = 0;
+  do
+    {
+      DigitalSet & S = shape.pointSet();
+      std::queue<DigitalSet::Iterator> Q;
+      for ( DigitalSet::Iterator it = S.begin(); it != S.end(); ++it )
+  if ( shape.isSimple( *it ) )
```
JacquesOlivierLachaud: Indentation looks weird on github here.

src/DGtal/topology/Object.ih
phcerdan: I can imagine! I digged directly into c++11 though. [Here](http://stackoverflow.com/questions/13816385/what-are-the-advantages-of-using-nullptr) says it solves overloading problems because nullptr is a pointer type, where 0 or NULL are int types.

JacquesOlivierLachaud:
Nice PR. Sorry for the long delay for reviewing it, I had several deadlines in the last weeks (SGP and ICPR). I check that everything compiles on my computer and the documentation and I merge.

src/DGtal/topology/NeighborhoodConfigurationsHelpers.h
phcerdan: TODO

src/DGtal/topology/Object.ih
JacquesOlivierLachaud: Yes I checked also. According to me, it is for mostly for the auto keyword, Overloading with int and pointer is kind of a weird usage.

phcerdan:
Thanks a lot for the reviewing and feedback, I am learning a lot with this library! I am right now contributing some segmentation filter in ITK, that I need for my thesis, but will try to improve this and the PR #1147 as soon as possible.

JacquesOlivierLachaud:
It is very nice for us to have contributors from far (far) away ! I was really surprised to see somebody from New Zealand who knows about isthmus and such stuff (so french :) )

JacquesOlivierLachaud:
Do you know Reinhardt Klette (he is in Auckland) ? He is one of the few guy in our community.

phcerdan:
No, I don't know him, I have the bad excuse that I am closer to Wellington. And yeah I know it is far away hehe, I am Spanish! Currently doing a PhD here in the antipodes, in a biophysics group , but I have been migrating to computer vision to characterize geometry/connectivity/topology of biopolymer networks

src/DGtal/topology/Object.ih
phcerdan: Maybe easier to move the function inside `Object.h`

JacquesOlivierLachaud: Do as is simplest for you !

JacquesOlivierLachaud:
Everything is fine for me and it works nicely on my computer. I wait for the few minor fixes (change of names mostly) and I merge.
And I think it will be time for you to go to bed, no ? :-)

phcerdan:
haha indeed. I will make the changes soon. Have a good afternoon!

src/DGtal/topology/CubicalComplexFunctions.h
```diff
@@ -437,6 +438,34 @@ namespace DGtal
                              BdryCellOutputIterator itBdry, 
                              InnerCellOutputIterator itInner );
 
+  /**
+   * Get the occupancy configuration of the neighborhood of a point in a cubical
+   * complex.
+   * The neighborhood size is considered DxD for dimension D of the point
```
JacquesOlivierLachaud: You mean 3^d in dD, so 3x3 in 2D and 3x3x3 in 3D.

JacquesOlivierLachaud:
Except from the above remark, it is fine for me to merge.

src/DGtal/topology/CubicalComplexFunctions.h
phcerdan: thx, changed.

JacquesOlivierLachaud:
Fine. I merge.

PR #1163, opened 2016-05-03
Fixing arithmetic ring issue

jlevallois:
# PR Description

The aim of this PR is to remove arithmetic ring issue when we change the integer type of `SpaceND`/`KhalimskySpaceND`.
By convention, we use `SpaceND<dim, int32_t>` and `KhalimskySpaceND<dim, int32_t>`.
But, in some case (if we want to reduce the memory footprint for example), we could be led to use `int16_t` instead of `int32_t` (or other).
The issue is that some algorithms are hard-coded to use `int32_t`. This PR try to resolve this issue.

*(I'm sorry, the reviewing process will not be easy because my editor has deleted some unnecessary space in comments ...)*

## Critical files modified

These files need to be closely reviewed.

- `AlphaThickSegmentComputer.h`
- `PowerMap.h` / `PowerMap.ih`
- `VoronoiMap.h` / `VoronoiMap.ih`
- `RigidTransformation3D.h`
- `Display2DFactory.ih`
- `DrawWithDisplay3DModifier.h`
- `BasicPointFunctors.h`

## Test files modified

- `testAlphaThickSegmentComputer.cpp`
- `testFrechetShortcut.cpp`
- `testFreemanChain.cpp`
- `testDistanceTransformationND.cpp`
- `testImageSimple.cpp`
- `testImageSpanIterators.cpp`
- `testSliceImageFromFunctor.cpp`
- `testDigitalSet.cpp`
- `testHyperRectDomain-snippet.cpp`
- `testHyperRectDomain.cpp`
- `testCellularGridSpaceND.cpp`
- `testSurfaceHelper.cpp`

# Checklist

- [N/A] Doxygen documentation of the code completed (classes, methods, types, members...)

:octocat:

### DIFF
--- a/src/DGtal/geometry/curves/AlphaThickSegmentComputer.h
+++ b/src/DGtal/geometry/curves/AlphaThickSegmentComputer.h
@@ -21,7 +21,7 @@
  * @author Bertrand Kerautret (\c kerautre@loria.fr )
  * LORIA (CNRS, UMR 7503), University of Nancy, France
  * @author Alexandre Faure
- * @author Fabien Feschet 
+ * @author Fabien Feschet
  * @author Mohammad Said
  * @author Jacques-Olivier Lachaud
  *
@@ -91,7 +91,7 @@ namespace DGtal
  * TInputPoint > .
  *
  *
- * This class is a model of boost::ForwardContainer and  CForwardSegmentComputer.  
+ * This class is a model of boost::ForwardContainer and  CForwardSegmentComputer.
  * It is also default constructible, copy  constructible, assignable and equality comparable.
  *
  *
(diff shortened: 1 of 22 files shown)


src/DGtal/geometry/curves/AlphaThickSegmentComputer.h
```diff
@@ -130,89 +130,89 @@ class AlphaThickSegmentComputer
...
   typedef typename InputPointContainer::size_type Size;
   typedef typename InputPointContainer::const_iterator ContainerConstIterator;
   typedef typename DGtal::functions::Hull2D::ThicknessDefinition ThicknessDef;
-  
+
   typedef typename InputPointContainer::iterator Iterator;
   typedef TConstIterator ConstIterator;
-  typedef ParallelStrip< SpaceND< 2,  DGtal::int32_t > ,true,true> Primitive;  
-  
+  typedef ParallelStrip< SpaceND< 2 > ,true,true> Primitive;
```
dcoeurjo: why did you change this one ? 
Even if int32_t is already the default type for SpacND, maybe the author as a good reason to want a int32_t type (if for instance we change the default type in ```SpaceND```

jlevallois: Agree, in this case I need to change `testAlphaThickSgementComputer` who uses `Z2i::Point`/`Z2i::Domain`

dcoeurjo: No I think it should be ````SpaceND<2, InputPoint::Coordinate>``` here

dcoeurjo:
nice PR BTW
gif-keyboard-11411730673171660395

src/DGtal/geometry/curves/AlphaThickSegmentComputer.h
jlevallois: We can't, because of `AlphaThickSegmentComputer< Z2i::RealPoint>` in `testAlphaThickSgementComputer` (line 199)

dcoeurjo: Ok I'll have a look.

jlevallois: It works when hardcoding `ParallelStrip<SpaceND< 2, int32_t>, ...>` because his `operator()` has a surcharge for `Point` and `RealPoint` (always double) from `SpaceND`.

dcoeurjo: Ok... And there is no implicit cast double -> int when used ?

jlevallois: `double` doesn't follow the concept `CInteger`, require by `SpaceND`

dcoeurjo: Yeah of course. I meant in the code, are input points casted to "primitive" points at some places (which would mean an implicit cast if input points are RealPoints)

dcoeurjo:
thanks again @jlevallois 
Can you please revert your edit in Alpha...  (reset the int32_t template type) and  add an entry to the changeling ?
Beside, the PR is fine, thanks again

jlevallois:
It's done :)

dcoeurjo:
Thanks Merging

PR #1170, opened 2016-05-09
some documentation module pages polished

troussil:
everything in the subject


dcoeurjo:
thanks for the edit. merging

PR #1177, opened 2016-05-15
fix rendering mode which was not applied on back faces

kerautret:
# PR Description
Small PR just to fix the Viewer3D rendering mode which was not activated for back faces.
By the way the double rendering was deactivated to rendering cube primitive which really improves performance when displaying large voxel set.

# Checklist
- [na] Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).
- [na] Doxygen documentation of the code completed (classes, methods, types, members...)
- [na] Documentation module page added or updated.
- [nn] New entry in the [ChangeLog.md](https://github.com/DGtal-team/DGtal/blob/master/ChangeLog.md) added.


dcoeurjo:
Thanks, merging

PR #1197, opened 2016-07-12
Fix eigen build

elcerdo:
*Thanks a lot for contributing to DGtal, before submitting your PR, please fill up the description and make sure that all checkboxes are checked. Please remove these lines in your PR.*

# PR Description

*your description here*

# Checklist

### DIFF
--- a/src/DGtal/math/linalg/EigenSupport.h
+++ b/src/DGtal/math/linalg/EigenSupport.h
@@ -52,6 +52,7 @@
 #include <iostream>
 #if defined(__GNUG__)
 #pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpragmas"
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #pragma GCC diagnostic ignored "-Wignored-attributes"
 #endif
--- a/src/DGtal/math/linalg/EigenSupport.h
+++ b/src/DGtal/math/linalg/EigenSupport.h
@@ -53,10 +53,12 @@
 #if defined(__GNUG__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#pragma GCC diagnostic ignored "-Wignored-attributes"
 #endif
 #if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdocumentation"
+#pragma clang diagnostic ignored "-Wignored-attributes"
 #endif
 #include <Eigen/Dense>
 #include <Eigen/Sparse>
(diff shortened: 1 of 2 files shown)


rolanddenis:
I confirm that it solves the compilation issue with Eigen 3.3 (Ubuntu 16.04) but I was wondering if it is better to replace `SparseMatrix::Index` by `SparseMatrix::StorageIndex` ([line 109](https://github.com/DGtal-team/DGtal/pull/1197/files#diff-1c9f5cfcf04fa25e3bcf437a59945657R109) ) instead of modifying default index for `SparseMatrix` typedef, following instructions in http://eigen.tuxfamily.org/index.php?title=3.3#Index_typedef .

Finally, it seems that this modification (replacing `Index` by `StorageIndex`) doesn't work for previous Eigen versions (even previous 3.x releases, as pointed out in https://forum.kde.org/viewtopic.php?f=74&t=132256#p355171 ) so OK for your solution.

The

Can I merge ?

dcoeurjo:
Before the merge, @elcerdo would have to add an entry to the change log. This bug fix might be critical for compatibility with Eigen.

dcoeurjo:
@elcerdo can you please add a changelog entry?

dcoeurjo:
Ping @elcerdo

elcerdo:
added changelog entry.
@rolanddenis i didn't experience the -Wshadow warnings. which version of eigen are you using?
i am using eigen 3.2.8 under archlinux.

src/DGtal/math/linalg/EigenSupport.h
```diff
@@ -52,11 +52,14 @@
 #include <iostream>
 #if defined(__GNUG__)
 #pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpragmas"
```
elcerdo: warnings can be ignored by adding a line here.

rolanddenis:
I get those warnings with `Eigen 3.3-beta1` (shipped with Ubuntu 16.04, shown as 3.2.92) and `Eigen 3.3-beta2` (shown as 3.2.93) with g++ (test with 5.4.0 and 4.9.3) in Debug mode and only `WITH_EIGEN` set (see https://gist.github.com/rolanddenis/65ecc38045e4fff22c54b4a12bf760e5 ).

No warnings when adding
```C++
#pragma GCC diagnostic ignored "-Wshadow"
```
or when using clang.

It's maybe related to this bug report http://eigen.tuxfamily.org/bz/show_bug.cgi?id=640 that also links to a gcc bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66472

dcoeurjo:
Okay.. Let's merge this one and check later the warnings. (I'll add a more recent gcc in the buildbot to keep track of such issues).

dcoeurjo:
thanks ;)

Issue #1200, opened 2016-08-06
compilation on debian gcc 4.7.2


kerautret:
A new error appears on Debian GNU/Linux 7 (wheezy)
g++ (Debian 4.7.2-5) 4.7.2
Copyright (C) 2012 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

/home/kerautre/DGtal/build/src/libDGtal.so: undefined reference to `DGtal::KhalimskySpaceND<3u, int>::NEG'
/home/kerautre/DGtal/build/src/libDGtal.so: undefined reference to `DGtal::KhalimskyPreSpaceND<2u, int>::NEG'
/home/kerautre/DGtal/build/src/libDGtal.so: undefined reference to `DGtal::KhalimskySpaceND<3u, int>::POS'
/home/kerautre/DGtal/build/src/libDGtal.so: undefined reference to `DGtal::KhalimskySpaceND<2u, int>::POS'
/home/kerautre/DGtal/build/src/libDGtal.so: undefined reference to `DGtal::KhalimskySpaceND<2u, int>::NEG'
/home/kerautre/DGtal/build/src/libDGtal.so: undefined reference to `DGtal::KhalimskyPreSpaceND<3u, int>::NEG'
/home/kerautre/DGtal/build/src/libDGtal.so: undefined reference to `DGtal::KhalimskyPreSpaceND<3u, int>::POS'
/home/kerautre/DGtal/build/src/libDGtal.so: undefined reference to `DGtal::KhalimskyPreSpaceND<2u, int>::POS'

kerautret:
Confirmed after upgrade to Jessie no more this errors...

rolanddenis:
While compiling DGtal library (default options ?) or an another program ?

kerautret:
yes it was compilling DGtal examples

rolanddenis:
Which one ? Any details ?
It could be a compiler bug but it is true that no reference can be taken to those static members if the class is not explicitly defined. I have tested with 4.7.4 (under Ubuntu 15.10) and examples compile successfully but not `testKhalimskySpaceND` (another error) :confused:

kerautret:
[...]

PR #1202, opened 2016-08-08
Fixing errors raised by g++ 4.7

rolanddenis:
# PR Description

Fixing following errors and warnings raised only by `g++ 4.7` ( 4.7.2 and 4.7.4 ):
- error with message `undefined reference to DGtal::KhalimskySpaceND<3u, int>::NEG` (same for `POS` and for `KhalimskyPreSpaceND`), see #1200. It is probably a bug of `g++` since there is no reason that he needs to take a reference to those static members (and there is no such errors with other compilers). However, adding a definition to those static members in namespace scope is a good thing since it will allow such usage if needed.
- ~~warning about unused parameter in defaulted constructors and assignment operators of `Khalimsky(Pre)SpaceND`. Don't know why he complains since it is only a definition but it's right that the parameter is useless here.~~ Eventually, I prefer to keep a proper documentation instead of clean compilation with `g++ 4.7`.
- error while constructing a `std::array` with initializer-list. A bug fixed in 4.8.0 : https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53361 .
- error with type aliasing, like in the following snippet:
```C++
using Image = ImageContainerBySTLVector<TDomain, TValue>;
using Value = typename Image::Value;
```
where it complains about not finding `Value` in `Image`. An another bug fixed in 4.8 : http://stackoverflow.com/questions/18213638/g-4-7-2-treatment-of-using-for-type-aliasing-seems-to-be-broken

It now compiles fine with `g++ 4.7.2` (on Debian Wheezy) and `g++ 4.7.4` (on Ubuntu 15.10) but the `C++11` support of 4.7.x versions is very buggy.

In addition, compilation with 4.8.5, 4.9.4, 5.2.1 on Ubuntu 15.10, and with 5.4.0 on Ubuntu 16.04 is done without warnings and errors with FFTW3, EIGEN (with #1197 applied), GMP, HDF5, OPENMP and QGLVIEWER options.

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,10 @@
 # DGtal 0.9.3
 
+## Bug Fixes
+
+- *Configuration/General*
+ - Fixing errors and warnings raised by g++ 4.7.x.
+   (Roland Denis, [#1202](https://github.com/DGtal-team/DGtal/pull/1202))
 
 # DGtal 0.9.2
 
--- a/src/DGtal/geometry/surfaces/estimation/estimationFunctors/SphericalHoughNormalVectorEstimator.h
+++ b/src/DGtal/geometry/surfaces/estimation/estimationFunctors/SphericalHoughNormalVectorEstimator.h
@@ -113,7 +113,7 @@ namespace DGtal
         SphericalAccumulator<RealPoint> accum(mySize);
         
         //We precompute the random rotations and accumulators
-        for(auto i=0; i < myNbAccumulators; ++i)
+        for(auto i = 0u; i < myNbAccumulators; ++i)
         {
           Matrix m = randomRotation();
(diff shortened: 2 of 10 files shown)


kerautret:
Nice, I will have a look when you think that ready

kerautret:
thank a lot!

rolanddenis:
It is ready but as you may have see, I have reverted the commit about the warnings on "unused parameters". It should compile fine with `g++ 4.7` but with warnings on every `Khalimsky(Pre)SpaceND` usage (or every class that uses the same syntax).

I will trigger again Travis since `clang` segfault during compilation but I think you can already take a look :wink:.

dcoeurjo:
Thanks a lot. Merging.

PR #1209, opened 2016-09-30
Voxelization

hadjiszs:
# PR Description
## Algorithm

I traverse all voxels V of triangle bounding box and insert in digital set if V overlaps current triangle.
I use intersection target ([samuli laine](https://mediatech.aalto.fi/~samuli/publications/laine2013egsr_paper.pdf#3)) to know if a voxel overlaps a triangle.
### 6-separated

6-separated

```
foreach triangle ABC
    foreach blue segment S of intersection target set
        2D projection of ABC on the plane of normal N(x,y,z)
        corresponding to S, just by ignoring coordinate

        foreach voxel V of ABC bounding box
            2D projection of V
            if 2D V is inside 2D triangle ABC
                insert V in digitalSet
```

### 26-separated

26-separated

It's the same algorithm but I do differently of 6-separated way for projections.

```
foreach triangle ABC
    foreach blue segment S of intersection target set
        compute e1 and e2 to have a base with each normal N(x,y,z)
        corresponding to S, so
            e1 = (-y, x, 0)
            e2 = (-xz, -yz, x^2)

        2D projection of A B and C on plane (0, e1, e2)
        foreach voxel V of ABC bounding box
            2D projection of V on plane (0, e1, e2)
            if 2D V is inside 2D triangle ABC
                insert V in digitalSet
```
  
# Checklist
- [x] No warning raised in Debug `cmake` mode (otherwise, Travis C.I. will fail).


kerautret:
Thanks! and also to the tools I look it this week end!

copyme:
Can you provide information about the implemented algorithm?

dcoeurjo:
@kerautret and @copyme please hold on. @hadjiszs was a intern I had at Lyon and the code is not ready for review.

I asked @hadjiszs to PR to be able to discuss on his code.

kerautret:
ok let me know when ready but that nice idea  ;)

dcoeurjo:
@hadjiszs have you checked the « allow edits from maintainers » when you've created your Pull-Request ?

hadjiszs:
@dcoeurjo sorry I didn't know this option, but it seems checked

screenshot:
screenshot

dcoeurjo:
@hadjiszs : since we do not optimize the intersection tests for adjacent triangles, I suggest to remove the mesh from the constructor and add a "voxelise(Mesh)" method.
You could thus voxelise different meshes and get a unique digital set at the end. What do you think?

hadjiszs:
@dcoeurjo yes, with something like that?

```
MeshVoxelizer mv(...);

Mesh3 aMesh1(...);
Mesh3 aMesh2(...);
Mesh3 aMesh3(...);

mv.add(aMesh1);
mv.add(aMesh2);
mv.add(aMesh3);

mv.voxelize(); // to be able to have a parallelization on list of mesh in voxelize() method
```

by the way, I try to fix last build fail of travis, but I really don't understand error about `DGtal::HDF5Reader` https://travis-ci.org/DGtal-team/DGtal/jobs/167108444#L1111

dcoeurjo:
Even simpler:

````
MeshVoxeliser mv(domain);

mv.voxelise (aMesh1);
mv.voxelise(aMesh2);

Point a(...),b(...),c(...);

//add a unique triangle
mv.voxelise(a,b,c);

..
````

no need to keep the meshes (or a ref to) as members.

**Comment by dcoeurjo** [comment]
Don't bother with the HDF5 issue, I'll have a look.

thx

**Comment by hadjiszs** [comment]
sorry I haven't explained my example,
it was to have a parallel processing on list of meshes within voxelize method

something like that (in voxelize method)

```
#omp parallel for // just for example with openmp syntax
foreach mesh M
    // .. voxelization of M into digitalSet
```

To let programmer add/remove meshes in MeshVoxelizer, and he chooses when he wants to run parallel voxelization of all current list of meshes he established. With that, he can have a parallel voxelization of a set of meshes

but, there is no parallelism in MeshVoxelizer yet, so my proposition is useless for now and I think it would be eventually for later, and I will do like you said as of now

**Comment by dcoeurjo** [comment]
I've updated the interface. Your tool may be deprecated now.

File: src/DGtal/shapes/MeshVoxelizer.h [hunk 1]
```diff
@@ -109,15 +109,111 @@ namespace DGtal
...
+     * @tparam Dummy dummy parameter just to be able to specialize inside template class (FIXME with another location)
+     */
+    template<size_t Sep, size_t Dimension = 1, typename Dummy = void>
+    struct IntersectionTarget { };
+
+    /**
+     * Specialization for 6-separated with 1D intersection target
+     */
+    template<typename Dummy>
+    struct IntersectionTarget<6, 1, Dummy>
```
hadjiszs: I think I need to put `IntersectionTarget<>` implementation [(this block)](https://github.com/hadjiszs/DGtal/blob/d3f6b016a6a202ba6b33f83a17389ada01e6d24b/src/DGtal/shapes/MeshVoxelizer.h#L123-L209) out of `MeshVoxelizer<>` class, in a separate `.h/.ih` because here like that it seems not really ok and IntersectionTarget may be grow up in the future if 2d intersection target is implemented

Should I put this code in a separate .ih /.h ? And, in this case, where do you think I should put `"IntersectionTarget".(h|ih)` ? in the same module ? shapes ?

dcoeurjo: Agree with an external "intersectiontarget.h" file (same place). When defining the target, please consider integer only computations as far as possible.  (like 0.5-> 1 and "/2" if needed).

src/DGtal/shapes/MeshVoxelizer.h
```diff
@@ -109,15 +109,111 @@ namespace DGtal
...
+    {
+      const std::array<Edge, 3> myTarget {{
+          { { 0.5, 0.0, 0.0 }, { -0.5,  0.0,  0.0} } , // AB
+          { { 0.0, 0.0, 0.5 }, {  0.0,  0.0, -0.5} } , // CD
+          { { 0.0, 0.5, 0.0 }, {  0.0, -0.5,  0.0} }   // EF
```
dcoeurjo: like here

src/DGtal/shapes/MeshVoxelizer.h
```diff
@@ -0,0 +1,321 @@
...
+          { {  0.5, 0.5, -0.5}, { -0.5, -0.5,  0.5} } , // CE
+          { { -0.5, 0.5, -0.5}, {  0.5, -0.5,  0.5} }   // DF
+        }};
+
+      std::array<VectorR3, 4> myE1base;
+      std::array<VectorR3, 4> myE2base;
+
+      IntersectionTarget<26, 1, Dummy>()
+      {
+        double coef = 1. / std::sqrt(3.);
```
dcoeurjo: ok my 0.5 vs. 1/2 comment does not work here..

src/DGtal/shapes/MeshVoxelizer.h
hadjiszs: sorry I'm not really good about floating-point arithmetic, I didn't understand this point.
If I put `1/2` instead of `0.5` here, it will be a integer division, so 0, then it will be cast to a double, so `0.0`, and I want `0.5`, no?

dcoeurjo:
hi @hadjiszs , Thanks for your commits,  Can I review this PR ?
Have you also updated the DGtalTools PR with your new edits ?

hadjiszs:
hi @dcoeurjo , yes I think it's ready for a new review.
codacy think some functions are never used, but they are indeed used, I think it's just because these functions are on scopes behind using declarations and so on.

Nothing to update about the tool, because there are no changes to the interface of MeshVoxelizer module, I've try with the current version of MeshVoxelizer and it seems ok.

copyme:
@hadjiszs @dcoeurjo I do have a question about the method. Can it handle non-convex meshes and meshes of complex topologies? I was working during my bachelor on some problems and it was not that easy to get a correct voxelization of some meshes. See for instance, [Rys 5., p 162](https://perso.esiee.fr/~plutak/files/MIS2012.pdf), where we showed some problems with vtkPolyDataToImageStencil method applied to a tree like mesh -- as you see the result is far from being correct. I ended up by voxelizing my trees branch-by-branch. If I remember correctly even tools like [binvox](http://www.patrickmin.com/binvox/) (Fakir Nooruddin and Greg Turk algorithm) had sometimes problems with complex meshes. I wonder if you aimed to solve some well-known issues.

dcoeurjo:
The approach is defined on a "per triangle" basis (and we only digitize the surface, not the interior). As long as you can triangulate your faces (easy), the digitization would be perfect. 
I've also experienced some issues with binvox when you use the z-buffer/OpenGL approach (but no issue with the "exact" mode)... but there's no such issue  here

copyme:
@dcoeurjo how about 3D curves, do you think your method could---after some modifications---handle them too? When I was working on L-MST 3D I tried to test it with some digitized curves and I have noticed some problems related to the curvature and grid. I mean, I do have some homemade code to digitized parametric curves but this does not always work.

dcoeurjo:
@hadjiszs can you please have a look to the ```box6-dig.obj``` when running the ```textMeshVoxelization``` ?
I'm a bit concerned by the extra-voxels at the corners...

dcoeurjo:
thx @hadjiszs

dcoeurjo:
a module doc is missing but the code can be reviewed

dcoeurjo:
Doc ok, this PR can be reviewed ;)

dcoeurjo:
can someone review this dude too ?

src/DGtal/shapes/doc/moduleMeshVoxelization.dox
```diff
@@ -0,0 +1,94 @@
...
+
+@snippet examples/shapes/exampleMeshVoxelizer.cpp MeshVoxelizerDig
+
+Once exported
+
+@snippet examples/shapes/exampleMeshVoxelizer.cpp MeshVoxelizerExport
+
+we obtain the following voxel set:
+
+@image html resultCube.png "Resulting voxelSet (quad faces triangulated by the viewer)"
```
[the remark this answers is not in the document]
kerautret: if you use an imported mesh in the examples simply use:
```
#include <DGtal/io/readers/MeshReader.h>
aMesh << "bunny.off";
```
and then simply output in sdp format and display it with 3dSDPViewer  for instance.

src/DGtal/shapes/MeshVoxelizer.h
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * Voxelize the mesh into the digital set.
+     * @warning if the mesh has non-trianuglar faces, we naively triangulate
+     * using a triangle fan at zero. If the face is not convex, the output
+     * will not be correct.
+     *
+     * @param [out] outputSet the set that collects the voxels.
+     * @param [in] aMesh the mesh to voxelize (vertex coordinates will
+     * be casted to @e PointR3 points.
+     * @param [in] scaleFactor the scale factor to apply to the mesh
```
kerautret: for this parameter perhaps you have to alter to be careful to not use a value compatible with the domain of the digital set, else we have an error... (or/and detect it explicitly later ?)

src/DGtal/shapes/MeshVoxelizer.h
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Voxelize a unique triangle (a,b,c) into the digital set.
+     * voxels are inserted to the @e outputSet.
+     *
+     * @param [out] outputSet the set that collects the voxels.
+     * @param [in] a the first point of the triangle
+     * @param [in] b the second point of the triangle
+     * @param [in] c the third point of the triangle
+     * @param [in] scaleFactor the scale factor to apply to the triangle (default=1.0)
```
kerautret: same here

src/DGtal/shapes/MeshVoxelizer.ih
```diff
@@ -0,0 +1,241 @@
...
+
+          // if target is on the same side -> skip current iteration
+          if( isSameSide )
+            continue;
+
+          PointR2 pp = myIntersectionTarget.project(i, v);
+
+          // check if current voxel projection is inside ABC projection
+          if(pointIsInside2DTriangle(AA, BB, CC, pp) != OUTSIDE)
+            outputSet.insert(v);
```
kerautret: here I got an error if the scaleFactor is too large (v is outside the output domain, see previous remarks).

src/DGtal/shapes/MeshVoxelizer.ih
```diff
@@ -0,0 +1,241 @@
...
+{
+  std::pair<PointR3, PointR3> bbox;
+  VectorR3 n, e1, e2;
+  PointR3 A, B, C;
+
+  //Scaling + casting to PointR3
+  A = a*scaleFactor;
+  B = b*scaleFactor;
+  C = c*scaleFactor;
+
```
kerautret: perhaps add an assert to check if the points are in the digital set domain ?

src/DGtal/shapes/MeshVoxelizer.h
dcoeurjo: what do you mean ?
That the scale may induce voxels outside the set domain ?

src/DGtal/shapes/MeshVoxelizer.ih
dcoeurjo: yeah thanks.

dcoeurjo: Ok I've slightly changed the interface: I first check that the voxel is inside the domain or not before the insert. Hence, no more issue with the domain but if a voxel is outside the set, it is simply skipped.
Fine for you ?

dcoeurjo:
Thanks for your comments.

dcoeurjo:
Could you please generate the images with the bunny and 2 digitizations using different resolution (I'm not familiar with the mesh viewer) ? 
This would be very nice for the module doc.

src/DGtal/shapes/MeshVoxelizer.ih
kerautret: yes fine

kerautret:
> Could you please generate the images with the bunny and 2 digitizations using different resolution (I'm not familiar with the mesh viewer) ?
This would be very nice for the module doc.

yes is what you want ?
bunnys250
bunnys500

src/DGtal/shapes/MeshVoxelizer.h
kerautret: sorry something was auto completed with my editor, or I was very tired ;)
I was thinking that you could add a  user warning  in the doc to avoid this potential error. But with the last change, we can simply mention that outside domain voxel are ignored ?

dcoeurjo:
thx @kerautret

dcoeurjo:
I've updated the doc with the bunny images. Would it be fine for you ?

dcoeurjo:
@kerautret is it fine for you?

src/DGtal/shapes/doc/moduleMeshVoxelization.dox
```diff
@@ -84,6 +84,11 @@ be automatically triangulated using a naive fan approach. Hence, if
 the face in not planar and/or convex, the resulting digitization might
 be incorrect.
 
+Here you have another example with a more complex mesh (using DGtaltools tools for visualization):
+
+@image html bunnies.png "Voxelization at different resolution of a Stanford Bunny."
```
kerautret: different resolutionS ?

dcoeurjo: damit ;)

kerautret:
thanks looks fine, just waiting for travis and merging....

dcoeurjo:
travis done ;)
merging

PR #1222, opened 2016-12-08
Avoid enforcing std=c++11 when not necessary.

### LINKED ISSUE #1219: Issue with DGtalConfig.cmake
Using FIND_PACKAGE(DGtal) sets -std=c++11, without warning the user.
He cannot for example ask for -std=c++14 before using FIND_PACKAGE, as it will be "overridden" by calling FIND_PACKAGE.

phcerdan:
# PR Description
Avoid enforcing flag `-std=c++11` when not necessary.
For example when the default flag of the compiler is
-std=gnu++11,c++14, etc... Or when the user explicitly set a std flag.

fix #1219
 
# Checklist

Tested with gcc 6.2 that sets std=gnu++14 by default, and with gcc 5.4 that doesn't set any std flag.


phcerdan:
I've just found this in cmake: https://cmake.org/cmake/help/v3.0/module/CheckCXXCompilerFlag.html

dcoeurjo:
Thanks @phcerdan for the PR. Your edits look fine but I also think that using ```CheckCXXComplilerFlag``` could be a better solution than the adhoc ```try``` we have..

phcerdan:
It does seem that CheckCXXCompilerFlag, and [CheckCXXSourceCompiles](https://cmake.org/cmake/help/v3.0/module/CheckCXXSourceCompiles.html?highlight=checkcxxsourcecompiles), are wraps around try_compile.

Also there is a new CXX_STANDARD variable, but it does not get passed to the CheckCXX... . See cmake bug: https://gitlab.kitware.com/cmake/cmake/issues/16456

So I think the existing try_compile is not that bad until cmake API stabilizes.

dcoeurjo:
Ok thanks

dcoeurjo:
Could you please add an entry to the change log ?

dcoeurjo:
Maybe we also should fix the DGtalConfig.cmake.in to re-discover cpp11

phcerdan:
DGtalConfig.cmake modified using variable defined in CheckCPP11.cmake: `@CPP11_COMPATIBLE_BY_DEFAULT@`.
Tested with `gcc6.2` (`-std=gnu++14`) and with `gcc5.4` (no `std` flag) and it seems to work as expected, ie, avoid setting any `std` flag for the first case, and setting `std=c++11` for the second.

dcoeurjo:
Excellent ! Thanks a lot.

dcoeurjo:
All good. merging. thanks again

PR #1230, opened 2017-01-03
Installing tables now reads ENV{DESTDIR} variable.

### LINKED ISSUE #1229: make install DESTDIR="<path>" ignored by some files
Since version 0.9.2 there is a problem with installing DGtal to a directory pointed by DESTDIR. For example one can notice that there is an attempt to copy into
/usr/include/DGtal/topology/tables/NeighborhoodTables.h.tmp, instead to a directory pointed by DESTDIR.

The problem seems to be related to the look up tables introduced in 0.9.2. The issue was extracted from  #1207.

phcerdan:
# PR Description
`make DESTDIR=/home/joe install` didn't work for installing LU tables. This approach is used for linux packagers. Fix #1229

CMake uses DESTDIR at install time, but because the installation of the LU tables uses manual `CODE`, CMake does not prepend DESTDIR to `CMAKE_INSTALL_PREFIX` in this case. At least that is my interpretation from:  https://gitlab.kitware.com/cmake/cmake/issues/11704#note_212469

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,7 +9,8 @@
  - The project has a now a unique compiled library: DGtal. The DGtalIO
    target has been removed. (David Coeurjolly,
    [#1226](https://github.com/DGtal-team/DGtal/pull/1226))
-
+ - Installation of tables is able to use DESTDIR env variable if provided by user.
+   (Pablo H Cerdan, [#1230](https://github.com/DGtal-team/DGtal/pull/1230))
 
 - *Geometry Package*
  - VoronoiMap, PowerMap, (Reverse)DistanceTransformation and ReducedMedialAxis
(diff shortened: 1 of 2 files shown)


phcerdan:
`travis` failing to find `qt4` in one mac build with clang? https://travis-ci.org/DGtal-team/DGtal/jobs/188695861

dcoeurjo:
Yes Qt 4 is no longer supported on MacOS Sierra. We need to force Qt5 in Travis.

dcoeurjo:
Travis fix is here: #1231

dcoeurjo:
@phcerdan can you please pull the master again ? (linux bots have been canceled, fixed, now).

Should we wait for zlib (and compressed LUT) before merging this PR or I can review it right now ?

phcerdan:
@dcoeurjo rebased on top of master.
This fixes a current issue, even though we will change how we deal with tables using `zlib`. I think it is ready to review.

phcerdan:
#1233 has another approach to this, instead of using the `install(CODE ...` that has to manually manage `DESTDIR`, #1233 generates a temporary configure_file, and then install it.

So I would recommend to not use this if we are merging #1233 soon.

Issue #1221, opened 2016-12-06
std::binder2nd is deprecated in c++11


phcerdan:
Remove typedef and replace with a lambda?
Or if IntPredicate has to be stored: decltype(std::bind)?
http://stackoverflow.com/questions/19772691/what-replaces-binder2nd-in-c11

```bash
[20%] Building CXX object tests/kernel/CMakeFiles/testImagesSetsUtilities.dir/testImagesSetsUtilities.cpp.o
/home/phc/Software/DGtal/src-fork/tests/kernel/testPointPredicateConcepts.cpp: In function ‘int main(int, char**)’
/home/phc/Software/DGtal/src-fork/tests/kernel/testPointPredicateConcepts.cpp:99:16: warning: ‘template<class _Operation> class std::binder2nd’ is deprecated [-Wdeprecated-declarations]
   typedef std::binder2nd< std::less<int> > IntPredicate;
                ^~~~~~~~~
In file included from /usr/include/c++/6.2.1/bits/stl_function.h:1127:0,
                 from /usr/include/c++/6.2.1/string:48,
                 from /usr/include/c++/6.2.1/bits/locale_classes.h:40,
                 from /usr/include/c++/6.2.1/bits/ios_base.h:41,
                 from /usr/include/c++/6.2.1/ios:42,
                 from /usr/include/c++/6.2.1/ostream:38,
                 from /usr/include/c++/6.2.1/iostream:39,
                 from /home/phc/Software/DGtal/src-fork/tests/kernel/testPointPredicateConcepts.cpp:31:
/usr/include/c++/6.2.1/backward/binders.h:143:11: note: declared here
     class binder2nd
           ^~~~~~~~~
/home/phc/Software/DGtal/src-fork/tests/kernel/testPointPredicateConcepts.cpp:101:16: warning: ‘template<class _Operation> class std::binder2nd’ is deprecated [-Wdeprecated-declarations]
   typedef std::binder2nd< std::greater<float> > FloatPredicate;
                ^~~~~~~~~
In file included from /usr/include/c++/6.2.1/bits/stl_function.h:1127:0,
                 from /usr/include/c++/6.2.1/string:48,
                 from /usr/include/c++/6.2.1/bits/locale_classes.h:40,
                 from /usr/include/c++/6.2.1/bits/ios_base.h:41,
                 from /usr/include/c++/6.2.1/ios:42,
                 from /usr/include/c++/6.2.1/ostream:38,
[...]

PR #1232, opened 2017-01-09
Fix missing template parameter for overloaded functions that use PointVector.

rolanddenis:
:tada: :beers: **Happy New Year to all the DGtal team !** :beers: :tada:

# PR Description

The hint comes from a compilation test of DGtal using Intel Compiler 17. It raises errors in display-related classes (`Display[23]DFactory`, `Viewer3DFactory`, ...) about overloaded functions/methods not found.

A minimal example is:
```C++
template < unsigned int N >
struct Container {};

template < int N, typename TContainer = Container<N> >
struct Dummy {};

template < int N >
void f( Dummy<N> const & )
{
}

int main()
{
    Dummy<2> toto;
    f(toto);
    return 0;
}
```
that fails to compile with error:
```
test.cpp(15): error: no instance of function template "f" matches the argument list
            argument types are: (Dummy<2, Container<2U>>)
      f(toto);
      ^
test.cpp(8): note: this candidate was rejected because at least one template argument could not be deduced
  void f( Dummy<N> const & )
```

The problem comes from the different integer type in template parameters (`int` versus `unsigned int`) that leads `icpc` to ignore the default `TContainer` type when searching for overloaded functions/methods.

In DGtal, it corresponds to the `Dimension` typedef that is a signed integer versus the unsigned `std::size_t` that is used to specify the size of a `std::array` (the default container for `PointVector`).

It is probably a bug since `g++` and `clang++` doesn't matter about this issue, **but anyway**, the `TContainer` template parameter is actually missing from those overloaded functions/methods in order to accept `PointVector` with non-standard value container, thus leading to this PR submission.

Intel Compiler 17 now compiles without error, but still some warnings about deprecated use of `std::binder2nd` (see #1221) and another warning about an invalid usage of `auto` (probably another bug). Version 16 still fails to compile DGtal (default options) but on a clearly valid `C++11` syntax.

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -30,6 +30,9 @@
    (I. Sivignon, [#1187](https://github.com/DGtal-team/DGtal/pull/1187))
 
 - *IO Package*
+ - Missing TContainer template parameter for overloaded functions/methods that
+   rely on PointVector.
+   (Roland Denis, [#1232](https://github.com/DGtal-team/DGtal/pull/1232))
  - Viewer3D: fix bad rendering when changing the scale.
    (Bertrand Kerautret, [#1217](https://github.com/DGtal-team/DGtal/pull/1217))
 
--- a/src/DGtal/io/Display2DFactory.h
+++ b/src/DGtal/io/Display2DFactory.h
@@ -380,20 +380,20 @@ template < typename TKSpace, typename TCellContainer >
     
     
 // PointVector
-template<Dimension dim, typename TComponent>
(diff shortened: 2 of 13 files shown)


rolanddenis:
Where are all Travis jobs ? :scream:

dcoeurjo:
😄 my bad... While working on MacOs bots, I've disabled linux ones...

dcoeurjo:
and happy new year 🎉

dcoeurjo:
travis up and running ;) can you please merge the master again?

rolanddenis:
Done :octocat:

PR #1233, opened 2017-01-11
Use zlib and boost::streams to read compressed LUTs.

phcerdan:
# PR Description

Current behavior is to decompress tables at cmake level,
wasting HDD space of user.

Depends on PR #1228 that adds zlib dependency.
Discussion and motivation on Issue #1229.
Also leaves the fix #1230 obsolete..

# Checklist

- [x] Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).
- [x] Doxygen documentation of the code completed (classes, methods, types, members...)
- [ ] Documentation module page added or updated.
- [ ] New entry in the [ChangeLog.md](https://github.com/DGtal-team/DGtal/blob/master/ChangeLog.md) added.
- [ ] No warning raised in Debug ```cmake``` mode (otherwise, Travis C.I. will fail).
- [ ] All continuous integration tests pass (Travis & appveyor)

PR #1228, opened 2016-12-27
Compressed Vol format

dcoeurjo:
# PR Description

This PR adds a new Version (3) to Vol file format allowing to have compressed volumetric data using zlib.  In Version 2, vol consists in an ASCII header and then the raw data in binary mode. This new version compresses the raw data using zlib/boost-filter, saving  huge storage.

VolReader and VolWriter have been updated and can still import/export Version 2 vols. 

To be able to perform such compression, zlib must be installed in the system (and is thus a new mandatory dependency in DGtal). This lib is highly standard and shouldn't be a problem. I could have make this dependency optional but if you all agree, I would like this "compressed" version to be the default one.

I've planned a ```upgradeVol``` tool to upgrade a vol file from Version 2 to 3 in ```DGtalTools```.

Additionaly, the Boost lib filter I use needs boost>1.50.

🎄 🎁 🎄 

## TODO
 - [x] update LongvolReader/LongvolWriter accordingly
 - [x] Check DGtalConfig.cmake.in

# Checklist

- [x] Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).
- [x] Doxygen documentation of the code completed (classes, methods, types, members...)
- [x] Documentation module page added or updated.
- [x] New entry in the [ChangeLog.md](https://github.com/DGtal-team/DGtal/blob/master/ChangeLog.md) added.
- [x] No warning raised in Debug ```cmake``` mode (otherwise, Travis C.I. will fail).
- [x] All continuous integration tests pass (Travis & appveyor)

PR #1234, opened 2017-01-20
Fix configure_file installing tables

phcerdan:
Escaping \${TABLE_DIR} in configure_file not getting the right value when using DESTDIR:
`make -j4 DESTDIR=~/dgtal_install install`
Revert to hard type the content of TABLE_DIR inside configure_file.

Follow on #1233 #1228 

# Checklist


phcerdan:
Ok, I am confused.
cmake_install.cmake is created when you run cmake first time (even without make install).
`cmake_install.cmake` will have 
If after that first configure you run `make -j4 DESTDIR=~/dgtal_install install`

PR #1235, opened 2017-01-20
Fix DESTDIR behavior at install time.

### LINKED ISSUE #1229: make install DESTDIR="<path>" ignored by some files

phcerdan:
We have to read DESTDIR at install time, not at configure time.
so we escape the `$` symbol.
`\$ENV{DESTDIR}`

The same with TABLE_DIR inside configure_file.

Read more: https://cmake.org/pipermail/cmake-developers/2013-January/017810.html

Fix #1229 
# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -37,6 +37,8 @@
    (David Coeurjolly, [#1208](https://github.com/DGtal-team/DGtal/pull/1208))
  - Continuous integration Travis bots are now based on ubunutu/trusty containers.
    (David Coeurjolly, [#1227](https://github.com/DGtal-team/DGtal/pull/1208))
+ - Fix usage of DESTDIR at install time for linux packagers.
+   (Pablo Hernandez, [#1235](https://github.com/DGtal-team/DGtal/pull/1235))
 
 - *Geometry Package*
  - ArithDSSIterator: fix missing postfix ++.
--- a/cmake/NeighborhoodTablesConfig.cmake
+++ b/cmake/NeighborhoodTablesConfig.cmake
@@ -13,9 +13,11 @@ configure_file(
 
 # ------ Install Tree ------ #
 #--- Configuration of the src/topology/tables/NeighborhoodTables.h.in for the install tree. Save to tmp file.
-# dev note: The variables are expanded before they are passed to install(CODE). We have to scape TABLE_DIR inside configure_file if we want it to be updated with the value set inside the install script.
+# dev note: we escape \$ENV{DESTDIR} because we want DESTDIR to be read at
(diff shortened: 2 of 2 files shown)


copyme:
@phcerdan It is pity that this is not documented in CMake. I already mentioned about doc problems via issue tracker of CMake. I hope it is going to be improved soon.

phcerdan:
@copyme yeah, they should document that `install(CODE...` and `install(SCRIPT...` don't manage `DESTDIR` by themselves. Thanks for that.
The tricky part here is that `cmake` generates a `cmake_install.cmake` at configure time, that will control the future installation. If we just put `$ENV{DESTIR}` it will be transformed to a string literal in `cmake_install.cmake` with the value at configure time of `DESTIR` (usually empty). But if we want `cmake_install.cmake` to read at install time `$ENV{DESTIR}`, it should be escaped.

I am not sure how well documented the escaping behavior is. But use of `cpack` requires two escape layers, (so for cpack to read `\` it needs `\\\`), `install(CODE...)` requires one, etc...

copyme:
@phcerdan I came to a conclusion that it would be better to start a new issue. See https://gitlab.kitware.com/cmake/cmake/issues/16583

phcerdan:
@copyme commented [there](https://gitlab.kitware.com/cmake/cmake/issues/16583) as well.

copyme:
@phcerdan I saw your comment. Thanks for additional information!

dcoeurjo:
Hi @phcerdan & @copyme, 
We're preparing a DGtal release and this PR is the last one that could be included. Have you found a solution for the @copyme's packing scripts on linux ?
Can I review this PR ?

copyme:
@dcoeurjo I will make tests on build.opensuse.org, stay tuned ;-)

phcerdan:
Hey @dcoeurjo definitely. This PR (of one character `\` !) should fix current problem reading DESTDIR at install time. But @copyme tests might ensure that (thanks!)

copyme:
@dcoeurjo @phcerdan I give you green light guys :-)

dcoeurjo:
Excellent ! ;)

@phcerdan, could you please add an tiny entry to the changelog (bug fix section) ?

The main drawback is that the commit size would drastically increased ;)

phcerdan:
done (rebased)

dcoeurjo:
👏

PR #1238, opened 2017-01-28
[FIX] Let CMake handle DESTDIR instead.

phcerdan:
Install a temporary file in InstallFiles at build time, and then use regular
install(FILES), which takes into account any DESTDIR logic.

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -39,6 +39,8 @@
    (David Coeurjolly, [#1227](https://github.com/DGtal-team/DGtal/pull/1208))
  - Fix usage of DESTDIR at install time for linux packagers.
    (Pablo Hernandez, [#1235](https://github.com/DGtal-team/DGtal/pull/1235))
+ - Fix, let CMake handle DESTDIR instead of manual manipulation.
+   (Pablo Hernandez, [#1238](https://github.com/DGtal-team/DGtal/pull/1238))
 
 - *Geometry Package*
  - ArithDSSIterator: fix missing postfix ++.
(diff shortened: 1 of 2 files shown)


copyme:
@hpcerdan I will test this today in the afternoon.

copyme:
@phcerdan Tested and I got: "    "/home/abuild/rpmbuild/BUILD/DGtal-0.9.3/src/DGtal/topology/tables/simplicity_table4_8.zlib";"

phcerdan:
@copyme thanks for testing it. For the building tree that result seems ok. Did you get it installing it or building it? The temporary file that will be copied into the installation tree is generated in `your_build_folder/InstallFiles/NeighborhoodTables.h`.

dcoeurjo:
Can I review this PR ?
@phcerdan @copyme is it ok ?

thx

phcerdan:
@dcoeurjo Yes, I have tested locally and I think it is ok. But @copyme might confirm.

To test it:
in build folder: we will install it inside the build folder, in a destdir folder.
`make -j4 DESTDIR=${PWD}/destdir install`

and then check that the installed `NeighborhoodTables.h` has the right path: `${CMAKE_INSTALL_PREFIX}` default to `/usr` I think:
`vim ./destdir/usr/include/DGtal/topology/tables/NeighborhoodTables.h`

it should only contain the install prefix.

dcoeurjo:
Thanks @phcerdan

@copyme, Can I have the green light ?

dcoeurjo:
BTW, could you please update the changelog ?

phcerdan:
@dcoeurjo in 0.9.3?

dcoeurjo:
fine for me. Merging this PR.
@copyme could you please double-check your scripts ?
thanks everyone

PR #1247, opened 2017-03-06
Added the configuration file .clang-format to use with clang-format

nnormand:
# PR Description

This PR adds the configuration file `.clang-format` for `clang-format`. The standalone tool `clang-format` can be used to format C++. It is documented [here](http://clang.llvm.org/docs/ClangFormat.html#standalone-tool) and can be installed from the clang distribution ([clang homepage](http://clang.llvm.org)). Once installed, `clang-format` is fairly easy to use:
`clang-format file` (output on the standard output) or `clang-format -i file` (inline formatting of `file`) with `.clang-file` in the directory or a a parent directory of the input file.

The set of rules were derived from DGtal [coding style](https://github.com/DGtal-team/DGtal/blob/master/CONTRIBUTING.md#coding-style) and from the DGtal source code base. `clang-rules` rules are available [here](http://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options). Also note that `clang-format` doesn't address all DGtal coding style instructions (_e.g._ variable naming) but can help creating a reasonably formatted code base.

# Checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,109 @@
+# .clang-format
+#
+# Alignment
+# ---------
+#
+# Align:
+# - fonction parameters on the open bracket,
+# - trailing comments.
+# Don't align:
+# - consecutive assignments or declarations,
+# - operands of binary and ternary expressions.
+#
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignOperands: false
+AlignTrailingComments: true
+#
+# Short statements
+# ----------------
+#
+# Forbid short
+# - blocks
+# - case labels
+# - functions
+# - if statements
+# - loops
+# on a single line.
+#
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+#
+# Line breaks
+# -----------
+#
+# Insert line breaks:
+# - before braces (Allman style),
+# - after the function definition return type,
+# - after template declarations,
+# - before multiline string literals,
+# - before ternary operators,
+# - constructor initializers before commas and align the commas with the colon.
+#
+# Limit the length of a line to 80 characters.
+#
+# If the constructor initializers don't fit on a line, put each initializer on
+# its own line.
+#
+# Keep empty lines at the start of blocks.
+#
+AlwaysBreakAfterReturnType: All
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: true
+BreakBeforeBraces: Allman
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: true
+ColumnLimit: 80
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+KeepEmptyLinesAtTheStartOfBlocks: true
+#
+# Indentation
+# -----------
+#
+# Use 2 spaces for indentations. Don't further indent continued lines.  Outdent
+# access modifiers (e.g. public:). Indent in all namespaces.
+#
+AccessModifierOffset: -2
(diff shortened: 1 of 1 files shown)


.clang-format
```diff
@@ -0,0 +1,109 @@
...
+# - fonction parameters on the open bracket,
+# - trailing comments.
+# Don't align:
+# - consecutive assignments or declarations,
+# - operands of binary and ternary expressions.
+#
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignOperands: false
```
dcoeurjo: I do like this one;)

dcoeurjo: (when set to true)

dcoeurjo:
Thanks for the PR, I'm on it with @kerautret

.clang-format
```diff
@@ -0,0 +1,109 @@
...
+AllowShortFunctionsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+#
+# Line breaks
+# -----------
+#
+# Insert line breaks:
+# - before braces (Allman style),
+# - after the function definition return type,
```
dcoeurjo: I prefer to have the return type in line if we have space (clang Penalty.. stuff for the 80col autofill)

dcoeurjo:
I've slightly modified the format (cf my test file https://gist.github.com/dcoeurjo/918f84fe1151accf8fd6a9ae1af9afb8)

What do you think ?

Two issues I still see:

* the doxygen comment is not ok (missing ' ') when clamped to 80col
* it would be nice to have "<<" aligned but I don't know if it is possible.

dcoeurjo:
@nnormand would you have an idea for the two issues I have ? Are my edits ok for you?

nnormand:
hi @dcoeurjo

I fully agree with your changes, the code looks nicer. However, for some reason, formatting with these new rules tend to induce more changes than before.

~~~
git checkout 85a146aa249b592c8d4c4d1eabd4e1b79405ca4e
(c=0;for f in $(find . -name '*.h' -o -name '*.ih' -o -name '*.cpp');do d=$(leven "$(expand -4 $f|perl -pe 's/\h+$//')" "$(clang-format $f)" || echo 0);echo "$f: $d";c=$(($c+$d));done;echo "total: $c")|tee 85a146aa249b592c8d4c4d1eabd4e1b79405ca4e.clang-format.log
git checkout 10c06d2d9fba327c8a248c7e9efa6130756aaa4e
(c=0;for f in $(find . -name '*.h' -o -name '*.ih' -o -name '*.cpp');do d=$(leven "$(expand -4 $f|perl -pe 's/\h+$//')" "$(clang-format $f)" || echo 0);echo "$f: $d";c=$(($c+$d));done;echo "total: $c")|tee 10c06d2d9fba327c8a248c7e9efa6130756aaa4e.clang-format.log
~~~

I'm not sure I understand your first issue and I don't know if the second one can be addressed.

dcoeurjo:
For the changes w.r.t. to the existing code, I'm not so really worried about that. Most files do not comply with the CodingStyle or formatting rules. 

Let's create "the format we want" in this PR and we'll see later if we decide to apply it to the existing code (or just ask the contributors to comply for their new PRs).

dcoeurjo:
(nice pre-commit script: https://dx13.co.uk/articles/2015/4/3/Setting-up-git-clang-format.html)

dcoeurjo:
@JacquesOlivierLachaud , @kerautret  can you please have a look to this clang-format ?

kerautret:
@dcoeurjo I try but I have this error : (I miss something  ?)

YAML:14:24: error: invalid boolean
AlignAfterOpenBracket: Align
                       ^~~~~
Error reading /Users/kerautre/.clang-format: Invalid argument
Can't find usable .clang-format, using LLVM style

kerautret:
that looks fine expect for the boost program_options but we could deactivate it with locally with ```// clang-format off```

dcoeurjo:
ok thanks. Merging this one.

PR #1259, opened 2017-04-02
Extend viewer3 d

JacquesOlivierLachaud:
# PR Description

This PR provides a reasonnable and easy way to extend the QGLViewer-based viewer of DGtal. Indeed, one cannot derive directly from the Viewer3D class and override methods. We provide here an Extension base class, that users have to derive in order to create a new interface. A simple example is provided.
Note also that this PR fixes some typos in class Viewer3D.

# Checklist

- NA Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).


JacquesOlivierLachaud:
For codacy/pr, I think the issue raised are stupid (methods are not called because it is event programming).
I modify the dox comments for travis-ci.

dcoeurjo:
agree... the code coverage tests in codacy are a bit crappy

JacquesOlivierLachaud:
I don't understand why travis fails ? Is there a configuration problem ?

dcoeurjo:
there is a brew issue when installing the deps on macOS bots. Just try to relaunch the build (there is nothing we can do I think)

dcoeurjo:
Travis ok now.

kerautret:
yes nice I look it! for codacy agree we can désactivante some detection i think ;)

examples/io/viewers/viewer3D-11-extension.cpp
```diff
@@ -0,0 +1,132 @@
...
+      {
+        typedef Viewer::KSpace KSpace;
+        Point p = viewer.space().lowerBound();
+        Point q = viewer.space().upperBound();
+        Point d = q - p;
+        Point a( ( rand() % d[ 0 ] ) + p[ 0 ],
+                 ( rand() % d[ 1 ] ) + p[ 1 ],
+                 ( rand() % d[ 2 ] ) + p[ 2 ] );
+        viewer << a;
+        viewer << Viewer::updateDisplay;
```
dcoeurjo: In this example, the ```updateDisplay``` resets the camera settings .. which is a bit odd.. 
Maybe this could be fixed in the ```Viewer3D```, @kerautret, any idea?

src/DGtal/io/doc/moduleQGLExtension.dox
```diff
@@ -0,0 +1,85 @@
...
+ */
+namespace DGtal {
+
+/**
+ * @page moduleQGLExtension Extending the QGLViewer Viewer3D interface
+ *  
+ *
+ * This part of the manual describes how to extend the standard
+ * Viewer3D interface of DGtal. For instance, you wish to assign new
+ * reactions to keys or to mouse event, or you wish to display or
```
dcoeurjo: reactions -> handlers?

src/DGtal/io/doc/moduleQGLExtension.dox
```diff
@@ -0,0 +1,85 @@
...
+ * interface provides a set of empty virtual methods so you just have
+ * to override in order to change the behavior of the viewer. For
+ * instance, you just have to override
+ * Viewer3D::Extension::keyPressEvent to capture other key events. You
+ * can override all the following methods:
+ *
+ * - Viewer3D::Extension::keyPressEvent. This method may be
+ *   overloaded to capture other key events.
+ *
+ * - Viewer3D::Extension::drawWithNames. This method is useful for
```
dcoeurjo: interesting.. would you have an example of that ?

examples/io/viewers/viewer3D-11-extension.cpp
kerautret: yes, updateDisplay call updateList with the update of the boundingbox (looks natural in classical use) but with this nice extension is should be better to add a new something like ```Viewer::updateDisplayFix``` which is equivalent but with no update. 
Else you can simply equivalently call viewer.updateList(false); instead Viewer::updateDisplay; and you will have it ;)

kerautret: I applied an improvement to avoid to add new display options, @JacquesOlivierLachaud I PR it on your branch: it checked if the BB has changed and it changed camera position only if the BB changed.

src/DGtal/io/doc/moduleQGLExtension.dox
JacquesOlivierLachaud: It is used in the mini-interaction method, see for instance
https://github.com/DGtal-team/DGtal/blob/master/examples/io/viewers/viewer3D-10-interaction.cpp
I do not have a useful example here for overriding it, but it was natural to provide ways to override all the standard methods provided by libQGLViewer.

examples/io/viewers/viewer3D-11-extension.cpp
JacquesOlivierLachaud: @kerautret With your PR, do I have now to call viewer.updateList(false); instead of Viewer::updateDisplay ?

src/DGtal/io/doc/moduleQGLExtension.dox
JacquesOlivierLachaud: Ok, if you prefer.

examples/io/viewers/viewer3D-11-extension.cpp
kerautret: not necessary in your example the BB don't change so the camera will not move. And in other cases it is looks fine if the  BB scene change.

JacquesOlivierLachaud: Thanks, I fix David's comment and the PR will be ready.

dcoeurjo:
👌 
Can I merge, @JacquesOlivierLachaud  ?

kerautret:
fine also for me ;)

JacquesOlivierLachaud:
I think so !

dcoeurjo:
go... ;)

PR #1264, opened 2017-05-08
fix type in DigitalSetFromMap

kerautret:
# PR Description
small pr just a type name fix on MS when including DigitalSetFromMap.h


dcoeurjo:
Thanks. Merging

Issue #1284, opened 2017-10-30
[deprecated] dynamic_exception


phcerdan:
Many warnings when compiling with gcc 7.2:
```
/src/DGtal/geometry/curves/FP.ih:45:27: warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]
    const TIterator& ite ) throw( InputException )
```

rolanddenis:
Yep, we should instead use `noexcept` to specify functions that don't throw. I think that we can first remove all `throw`specifications ... I'm on it unless you or @dcoeurjo are already fixing this ?

dcoeurjo:
go ahead;)

rolanddenis:
In progress in #1285. @phcerdan, do you still have warnings ?

phcerdan:
Closing via #1285 , thanks @rolanddenis!

PR #1285, opened 2017-10-30
Fixing deprecated throw exception specifications and indentation warnings

rolanddenis:
# PR Description
- Fixing deprecated throw exception specifications and replacing `throw()` by `noexcept`
- Fixing indentation warnings with g++ 7.x

# Checklist


dcoeurjo:
I am a bit confused... Does the fix consist in removing all exceptions ? 

Maybe I  haven't understood properly the ```noexcept```. I have the feeling that we are loosing something here (e.g. the type of the exceptions that was explicit before).

rolanddenis:
Specifying which exception a function may throw has been deprecated in C++11 for many reasons (see https://stackoverflow.com/questions/13841559/deprecated-throw-list-in-c11 and http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0003r0.html ). I agree that we loose the knowledge of what exception we may have to catch but it seems to be very hard to guarantee in a generic code ...

Instead of that, we can specify that a function doesn't throw any exception with `noexcept` or `noexcept(true)` (see http://en.cppreference.com/w/cpp/language/noexcept_spec ), or that her exception specification depends on the result of an expression ( `noexcept( expression )` ) or on other function exception specification using the `noexcept` operator ( see http://en.cppreference.com/w/cpp/language/noexcept ) , e.g `noexcept(noexcept(otherfunction()))`.

If a function has the  `noexcept(true)` specifier, no exception handler is generated (maybe a performance gain ???) and if it still throws an exception, it result in the program termination.

phcerdan:
Answering here from #1284:
tested with default configuration (examples ON), and no warnings! :+1:

dcoeurjo:
thanks @rolanddenis for the feedbacks..

rolanddenis:
We can additionally document the exceptions with the `@throw` doxygen command ?

dcoeurjo:
Ok travis is fine now. Shouldn't we remove the ```DGtal::IOException``` and update the doc accordingly ?

rolanddenis:
What do you mean ? Removing all the exceptions ?

dcoeurjo:
nevermind;) my remark was odd
Merging this dude..
thx

Issue #1283, opened 2017-10-30
Update CMake version requirement and avoid setting CXX_CMAKE_FLAGS


phcerdan:
`cmake 3.1` added the `CXX_STANDARD_REQUIRED`. Check [more about this in SO.](https://stackoverflow.com/questions/37621342/cmake-will-not-compile-to-c-11-standard)

Not a good practice to hardcode `CXX_CMAKE_FLAGS` with recent `cmake`.
I have found these recommendations from Daniel Pfeifer quite helpful [video](https://www.youtube.com/watch?v=bsXLMQ6WgIk), [slides](https://github.com/boostcon/cppnow_presentations_2017/blob/master/05-19-2017_friday/effective_cmake__daniel_pfeifer__cppnow_05-19-2017.pdf)

dcoeurjo:
So you're suggesting this, right ?

    set(CMAKE_CXX_STANDARD 11) # C++11...
    set(CMAKE_CXX_STANDARD_REQUIRED ON) #...is required...
    set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11

phcerdan:
Yes, also: `cmake_minimum_required(VERSION 3.1)` for that to work.

phcerdan:
Reference to other source codes transition: https://github.com/open-source-parsers/jsoncpp/issues/548

dcoeurjo:
something like that  #1290 ?

phcerdan:
Closing via #1290, thanks @dcoeurjo!

PR #1290, opened 2017-11-03
Cmake flags and cpp11

dcoeurjo:
# PR Description

Following #1283 , fix the way we set the c++11 flags in cmake.

# Checklist


phcerdan:
There are hard-coded flags in [DGtalConfig.cmake.in](https://github.com/DGtal-team/DGtal/blob/master/cmake/DGtalConfig.cmake.in) too.

I think we can remove CPPcheck11 completely (I sent you a PR for your branch doing this).

dcoeurjo:
thanks for the PR @phcerdan. I agree that CPP11CHECK is a bit overkill now but at least it would warn the user that is compiler is not C++11 compliant before trying to build. Maybe we could just keep one simple ```try_compile``` for that. 

What do you think ?

dcoeurjo:
(thx for ```DGtalConfig.cmake.in```, fixed)

dcoeurjo:
@kerautret could you please propagate this to DGtalTools and DGtalTools-contrib ?

kerautret:
@dcoeurjo yes I look it!

kerautret:
done (but just cmake version check since as discussed the check is done first by DGtal)

phcerdan:
@dcoeurjo sounds good to me!

phcerdan:
We should test this option:
`set(CMAKE_CXX_EXTENSIONS OFF) #...without compiler extensions like gnu++11`

I have faced problems in `boost` without the `-std=gnu++11` flag (just using `-std=c++11`):

One of many errors triggered from boost:
```
/home/phc/repository_local/SGEXT/build-superbuild/dep-install/include/boost/math/special_functions/detail/igamma_large.hpp:501:7: note: use -std=gnu++11 or -fext-numeric-literals to enable more built-in suffixes
/home/phc/repository_local/SGEXT/build-superbuild/dep-install/include/boost/math/special_functions/detail/igamma_large.hpp:502:7: error: unable to find numeric literal operator ‘operator""Q’
       BOOST_MATH_BIG_CONSTANT(T, 113, 0.243619480206674162436940696707789943e-10),
```

I did a workaround about this after reading about the same issue in [CGAL](https://github.com/CGAL/cgal/issues/1685) by using `-std=gnu++14` with `gcc`. But using `gnu++11` should suffice.

dcoeurjo:
@phcerdan ok but we don't to manually set the flags as suggested in CGAL fix.
Should I juste remove the line with the "OFF" to extensions ?

phcerdan:
@dcoeurjo I think so, the `OFF` in the extensions should suffice. I guess cmake then uses `gnu++11`  instead of `c++11` when using `gcc`.

dcoeurjo:
👌

dcoeurjo:
Merging

PR #1311, opened 2018-03-02
new url ?

kerautret:
# PR Description
by trying to fix the DGtalTools doc update, I wondering if we need to update the path since it is no more on the new one...
# Checklist

### DIFF
--- a/.travis/publish_doc.sh
+++ b/.travis/publish_doc.sh
@@ -4,7 +4,7 @@ set -ev
 
 cd build/
 echo $HOME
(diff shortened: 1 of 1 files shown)


dcoeurjo:
mm.. no need I think. `/home`and `/home-projects`  are the same

kerautret:
> mm.. no need I think. /homeand /home-projects are the same
on the new server I don't see it, but perhaps because it is hidden for tc18 account.. but nothing for /home/tc18

dcoeurjo:
ok let's merge anyway

Issue #1334, opened 2018-07-10
Doxygen log not checked anymore during Travis job


rolanddenis:
The Doxygen log file (`doxygen.log`) is not checked about warnings anymore. It seems to be a path problem : the file is searched in the current folder instead of the `build` folder.

See the tests in #1333 and job output https://api.travis-ci.org/v3/job/402309209/log.txt

rolanddenis:
WIP in #1335

PR #1306, opened 2018-02-01
MagickWriter in generic writer

kerautret:
# PR Description
To follow the @dcoeurjo PR with the new MagickWriter. We can now export color image using generic writer. 

# Checklist
- [x] Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).
- [x] Doxygen documentation of the code completed (classes, methods, types, members...)
- [x] Documentation module page added or updated.
- [x] New entry in the [ChangeLog.md](https://github.com/DGtal-team/DGtal/blob/master/ChangeLog.md) added.
- [x] No warning raised in Debug ```cmake``` mode (otherwise, Travis C.I. will fail).
- [x] All continuous integration tests pass (Travis & appveyor)

PR #1335, opened 2018-07-11
Fixing documentation checks and updating Travis build

rolanddenis:
# PR Description

Fixing documentation checks that was not fully done (see #1334)

# Checklist

- [x] Fixing all doxygen warnings
- ~~[ ] Adding `@example` tag check~~


.travis/build_dec.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+$SCRIPT_BEGIN
+
+### Dec
+echo "Compile Dec in non parallel mode to save memory (to fix gcc internal compiler error(Killed))";
+
+cd "$BUILD_DIR"
+make exampleDiscreteExteriorCalculusChladni
+#make exampleDiscreteExteriorCalculusSolve
+#make exampleDECSurface
```
dcoeurjo: why these one have been commented ?

rolanddenis: It is a copy-paste from `.travis/main_build.sh`:
https://github.com/DGtal-team/DGtal/blob/8f3ca8c0ddae25e1be2b745fbb1c039e494b0b91/.travis/main_build.sh#L16
and it has been commented in 163a8357c6d4e2380c0fd4e2a0c21093c7a08e0f from #1306 but I don't know why (@kerautret  ?).

I will try to uncomment these lines :octocat:

rolanddenis:
About the `ChangeLog.md`, I've already add an entry (see https://github.com/DGtal-team/DGtal/pull/1335/files#diff-e705c8fadf1193ab59443a5e6c8cbe8b) :wink:

.travis/build_dec.sh
kerautret: It was to try to avoid the segfault or timeout issue of travis due to gcc problem (as discussed with Jaco in last meeting). Changing the moment where it was compiled fixed the memory issue (I suppose), but perhaps we have to remove this example from the travis test.

rolanddenis: In fact, these two examples are already built during the main build. This part is only for examples that need to be build alone in order to avoid compiler failures. I think we can remove these two lines ...

rolanddenis: Oups, didn't see your answer :wink:

kerautret: yes of course if they are in the main part, but it was commented because it was an alternative.
But this part is not robust in all cases.

kerautret: I was wondering ;)

dcoeurjo:
Oops.. sorry ;)
Thanks for the PR

Just a quick question where do the SCRIPTBEGIN/END come from ?

rolanddenis:
They are defined in `.travis/before_global.sh`:
https://github.com/DGtal-team/DGtal/blob/f879fc4f45a10df44d6bcdd1c2a5d8cc6563949f/.travis/before_global.sh#L13
The idea is to set and reset some flags useful for debugging Travis, like the `set -e` that was already set in some scripts before this PR and that make a script returning a non-zero value if any of its command fails. We can also add `v` to view each executed commands. The `SCRIPT_END` resets the flags (because some scripts always have errors, like `.travis/install_deps_macos.sh` and Travis also have a strange behaviour when the `e` flag is set) and resets the working directory.

Issue #1268, opened 2017-06-20
testLinearStructure (dec) fails with recent Boost versions.


rolanddenis:
When compiling `testLinearStructure` using Boost version `1.63.0` or `1.64.0`, it fails with the following message:
```bash
New Block [testing 3d operators]
  [dec | primal 0-cells <-> dual 3-cells (2) | primal 1-cells <-> dual 2-cells (5) | primal 2-cells <-> dual 1-cells (4) | primal 3-cells <-> dual 0-cells (1)]
  New Block [base operators]
    d0
[primal 0-form => primal 1-form (2 => 5)]
-1  1
 0  1
 0  1
-1  0
-1  0
    d2p
[dual 2-form => dual 3-form (5 => 2)]
-1  0  0 -1 -1
 1  1  1  0  0
    [ERR] Fatal Error - assertion (Eigen::MatrixXd(d0.myContainer) == d0_th) failed in void test_manual_operators_3d(): /scratch/denis/DGtal/tests/dec/testLinearStructure.cpp(391)
Abandon (core dumped)
```
The related code https://github.com/DGtal-team/DGtal/blob/master/tests/dec/testLinearStructure.cpp#L391 suggests that the primal 0-form should instead be:
```C++
d0_th <<
    -1, 0,
    -1, 0,
    -1, 1,
     0, 1,
     0, 1;
```
It is indeed the case when using Boost `1.58.0`.

Tested with GNU `5.4.1`, `6.2.0` and `7.1.1`, with Eigen `3.3.3` and `3.3.4`.

dcoeurjo:
Looks ok with the last  boost 1.75. 
Should we keep this issue open?

rolanddenis:
Still fails on my side with GCC 10.2.0, Eigen 3.3.9 and Boost 1.75.0 ... Does it work on your side ?

dcoeurjo:
Ok so the problem does not come from boost 1.75 itself, maybe a combo GCC10+boost1.75.
(clang+boost1.75 is fine).

rolanddenis:
Fails also with LLVM 10.0.0, Boost 1.75.0 and Eigen 3.3.9 (Ubuntu), strange...

rolanddenis:
[...]

PR #1352, opened 2018-10-16
Add extra exceptions for arithmetic and types operations

copyme:
This PR allows for extra exceptions in the debug mode (G++).

So, without these few lines we have all the test passing:

> 100% tests passed, 0 tests failed out of 193
> 
> Total Test time (real) = 448.64 sec
> 

But if we run the tests with the additional exceptions we get:

> 99% tests passed, 2 tests failed out of 193
> 
> Total Test time (real) = 451.77 sec
> 
> The following tests FAILED:
>          54 - testMultiStatistics (Failed)
>          86 - testLambdaMST3D (NUMERICAL)
> Errors while running CTest

The bugs related to LambdaMST3D are already fixed in #1339

It is not that bad only two tests failed (and one of mine). We can think about adding different types of extra exceptions but allowing all may cause problems with boost, for example.


copyme:
First it looks like there is another flaoting-point error see:

/home/travis/build/DGtal-team/DGtal/tests/dec/testHeatLaplace.cpp:125: FAILED:
due to a fatal error condition:
  SIGFPE - Floating point error signal

Second the thing does not work on OS X. I will have to turn it on only for Linux.

dcoeurjo:
Can you see the issue with `testMultiStatistics`? I don't know of to trace back the bug..

copyme:
@dcoeurjo OK, I will try to find a solution once I will have a bit of time.

dcoeurjo:
Thx.. Not easy to locate the actual bug with the SIGFPE signal which does not provide a lot of information

copyme:
@dcoeurjo Yup, I use GDB to stop at the signal and try to understand where is the problem. I think there are better techniques to deal with this. I will need to have a look at.

dcoeurjo:
I see.. thx anyway..

dcoeurjo:
can you please merge with the master ? Are all tests passing now ?

copyme:
@dcoeurjo They should, lets see what travis will tell us right now.

copyme:
@dcoeurjo One test is still failing. The fix is in a non-merged branch: Curves3D. I am working on it.

dcoeurjo:
If #1339 is merged, will the remaining test be fine?

dcoeurjo:
ping @copyme

copyme:
@copyme it should be fine I will merge master and we will see

copyme:
@dcoeurjo sorry for the delay, I have not seen your message. I just added a changelog entry, and if the tests go then you can merge, I think.

copyme:
@dcoeurjo, in fact, there are bugs in the tests. The fix discovered new issues. I will have a look at and try to fix them.

dcoeurjo:
👌

copyme:
@dcoeurjo I will need a hand to fix these issues.

So, in `testHeatLaplace.cpp` the problem is here:

` 
const double cut_locus = log( - log1p( t ) ) + 2.;
`
the value of ` - log1p( t )` is `-0.0953102` but `std::log` does not accept negative values:

> If the argument is negative, NaN is returned and FE_INVALID is raised. 

Since I do not understand the code well (and I have no time to) I need a piece of precise information on how to fix this issue.

copyme:
ping @cgurps

cgurps:
You can replace this line by

`const double cut_locus = 3.0;`

The value doesn't really mater for this test

copyme:
@cgurps thanks, fixed. @dcoeurjo before merging please review my fix to the bug in `SphericalHoughNormalVectorEstimator`

copyme:
Also, on my machine I get this:

/home/kacper/Projects/DGtal/cmake-build-debug/tests/dec/testLinearStructure
New Block [testing 3d operators]
  [dec | primal 0-cells <-> dual 3-cells (2) | primal 1-cells <-> dual 2-cells (5) | primal 2-cells <-> dual 1-cells (4) | primal 3-cells <-> dual 0-cells (1)]
  New Block [base operators]
    d0
[primal 0-form => primal 1-form (2 => 5)]
-1  1
 0  1
 0  1
-1  0
-1  0
    d2p
[dual 2-form => dual 3-form (5 => 2)]
-1  0  0 -1 -1
 1  1  1  0  0
    [ERR] Fatal Error - assertion (Eigen::MatrixXd(d0.myContainer) == d0_th) failed in void test_manual_operators_3d(): /home/kacper/Projects/DGtal/tests/dec/testLinearStructure.cpp(391)

Process finished with exit code 134 (interrupted by signal 6: SIGABRT)

But maybe this is Eigen version related?

rolanddenis:
It is related to recent Boost versions (see https://github.com/DGtal-team/DGtal/issues/1268).

copyme:
@rolanddenis thanks for the info!

copyme:
@dcoeurjo I am done here. Just pay attention to the changes mentioned above while merging.

dcoeurjo:
all good.. thanks.

PR #1356, opened 2018-10-31
Fix bug in ImageLinearCellEmbedder

JacquesOlivierLachaud:
*Thanks a lot for contributing to DGtal, before submitting your PR, please fill up the description and make sure that all checkboxes are checked. Please remove these lines in your PR.*

# PR Description

This PR fixes a bug in class ImageLinearCellEmbedder when the image was made of unsigned values (for instance, a grayscale image made of unsigned char). In this case, the difference y2-y1 may sometimes be negative but the type forbids the value to be negative. The solution is to first cast values to double, then make the difference.

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -51,6 +51,10 @@
   - Adding missing headers in some files of DEC.
     (Roland Denis, [#1349](https://github.com/DGtal-team/DGtal/pull/1349))
 
+- *Image*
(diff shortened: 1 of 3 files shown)


JacquesOlivierLachaud:
Seems to pass everything except to long to compile by travis.

dcoeurjo:
Perfect. Thanks.

PR #1364, opened 2018-11-23
Some fixes for DGtal for dummies

JacquesOlivierLachaud:
# PR Description

Small fixes in format of some files. Add a few missing methods in Triangulated and Polygonal Surface. Add colormaps to shortcuts and some fixes in TickedColorMap.

# Checklist

- [x] Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).
- [x] Doxygen documentation of the code completed (classes, methods, types, members...)
- [x] Documentation module page added or updated.
- [x] New entry in the [ChangeLog.md](https://github.com/DGtal-team/DGtal/blob/master/ChangeLog.md) added.
- [ ] No warning raised in Debug ```cmake``` mode (otherwise, Travis C.I. will fail).
- [ ] All continuous integration tests pass (Travis & appveyor)

PR #1365, opened 2018-11-27
square norm in pointvector

dcoeurjo:
# PR Description

# Checklist

### DIFF
--- a/src/DGtal/kernel/PointVector.h
+++ b/src/DGtal/kernel/PointVector.h
@@ -736,9 +736,9 @@ namespace DGtal
      * from the type T to double for each components to compute the
      * norms.
      *
-     * @return the square norm of the point/vector as a double.
+     * @return the squared norm of the point/vector as a double.
      */
-    double sqnorm( ) const;
+    double squaredNorm( ) const;
 
     
     /**
--- a/src/DGtal/kernel/PointVector.h
+++ b/src/DGtal/kernel/PointVector.h
@@ -731,6 +731,17 @@ namespace DGtal
     double norm( const NormType type = L_2 ) const;
 
     /**
+     * Computes the square L2 norm of a point/vector.
+     * \warning This method performs a conversion
+     * from the type T to double for each components to compute the
+     * norms.
+     *
+     * @return the square norm of the point/vector as a double.
+     */
+    double sqnorm( ) const;
+
+    
+    /**
      * Computes the 1-norm of a vector.
      *
      * @return the absolute sum of the components of this vector.
--- a/src/DGtal/kernel/PointVector.ih
+++ b/src/DGtal/kernel/PointVector.ih
@@ -667,14 +667,11 @@ inline
 double
 DGtal::PointVector<dim, TComponent, TContainer>::sqnorm () const
 {
-  double tmp = 0.0;
   ASSERT ( dim > 0 );
+  double tmp = 0.0;
   double v;
   for ( DGtal::Dimension i = 0; i < dimension; i++ )
-  {
-    v =NumberTraits<Component>::castToDouble(myArray[ i ]);
-    tmp += v*v;
-  }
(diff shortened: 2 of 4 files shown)


src/DGtal/kernel/PointVector.ih
```diff
@@ -665,6 +665,22 @@ negate()
...
 double
+DGtal::PointVector<dim, TComponent, TContainer>::sqnorm () const
+{
+  double tmp = 0.0;
+  ASSERT ( dim > 0 );
+  double v;
+  for ( DGtal::Dimension i = 0; i < dimension; i++ )
+  {
+    v =NumberTraits<Component>::castToDouble(myArray[ i ]);
+    tmp += v*v;
```
rolanddenis: Why not simply `tmp += std::pow( NumberTraits<Component>::castToDouble(myArray[ i ]), 2);` ?

dcoeurjo: well.. for performances.. 
I expect that all implementations of `std::pow(...,2)` to be specialized and optimized  but I'm not sure so I prefer the code to be explicit.

for exponents !=2, I would have used the `std::pow`

rolanddenis: I think it should be optimized by any decent compiler (see https://godbolt.org/z/TYR7wA) but ok :wink:

dcoeurjo: :)

dcoeurjo:
oops.. this PR also contains @JacquesOlivierLachaud's commits on #1364

src/DGtal/kernel/PointVector.h
```diff
@@ -730,6 +730,17 @@ namespace DGtal
...
 
+    /**
+     * Computes the square L2 norm of a point/vector.
+     * \warning This method performs a conversion
+     * from the type T to double for each components to compute the
+     * norms.
+     *
+     * @return the square norm of the point/vector as a double.
+     */
+    double sqnorm( ) const;
```
JacquesOlivierLachaud: Is there a reason for choosing the name "sqnorm" instead of "squaredNorm" or "squaredL2Norm" ? Because it is not our usual naming conventions.

dcoeurjo: i've seen this name in previous lib.. But I can switch to squaredNorm is you want

dcoeurjo: let's go with `squaredNorm`

JacquesOlivierLachaud:
All good. Merging.

src/DGtal/kernel/PointVector.ih
rolanddenis: @dcoeurjo I was partly mistaken on that point: I knew that since C++11, all `std::pow` specializations (e.g. for `double^int`) were grouped into (almost) one remaining definition (see (7) in https://en.cppreference.com/w/cpp/numeric/math/pow) but I only realized very recently that all integral parameters were converted to floating-point type before applying the power operator (see the description of (7) on the same link).

Thus, when calculating `std::pow(v, 3)` with `v` a double, the exponent 3 is stored as a double (or converted to a double if unknown) before calling (and not inlining) the (thus general) `pow` function (see https://godbolt.org/z/3MwAqz and compare the result when using `-std=c++98`).

If working with `float`, things get worst: there exists a `std::pow(float, float)` overload (item (1)) but probably because passing the integer 3 as the exponent involves a conversion, the compiler choose the overload (7) (that use template parameters) and that leads to `float <-> double` conversions before and after calling the `pow` function (see https://godbolt.org/z/Rx-bvo). The right overload can be forced by passing `3.f` as exponent or by explicitly using the `std::powf` function.

However, it seems that with an exponent of 2, the compilers still inline the `pow` function without  spurious conversions but I don't know if it is in the standard...

PS: If you wonder about vectorization, look at https://godbolt.org/z/HRcsRW ...

PR #1366, opened 2018-11-28
Fixing warning of Clang when including GraphicsMagick 1.3.31

rolanddenis:
# PR Description

Fixing warning of Clang when including GraphicsMagick 1.3.31, so that Travis job on Mac doesn't fail.

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,8 @@
     [#1326](https://github.com/DGtal-team/DGtal/pull/1326)
   - Fixing documentation checks and updating Travis scripts
     (Roland Denis, [#1335](https://github.com/DGtal-team/DGtal/pull/1335))
+  - Fixing warning of Clang when including GraphicsMagick v1.3.31
+    (Roland Denis, [#1366](https://github.com/DGtal-team/DGtal/pull/1366))
 
 - Miscellaneous
   - Fix Small bug in Integral Invariant Volume Estimator in 2D
--- a/src/DGtal/io/readers/MagickReader.h
+++ b/src/DGtal/io/readers/MagickReader.h
@@ -45,9 +45,22 @@
 #include <cstdio>
 
 #if defined(WITH_MAGICK)
-#include <Magick++.h>
+// Specific inclusion method to fix warning with GraphicsMagic 1.3.31 and Clang.
+// The warning comes from two "diagnostic pop" with missing corresponding push.
+// "MagickLibAddendum" is defined in ImageMagick since 9 years but not in GraphicsMagic.
+#  if !defined(MagickLibAddendum) && defined(__clang__)
+#    pragma clang diagnostic push
+#    pragma clang diagnostic push
+#    include <Magick++.h>
+#    if MagickLibVersion != 0x221900
+#      pragma clang diagnostic pop
+#      pragma clang diagnostic pop
+#    endif
+#  else
+#    include <Magick++.h>
+#  endif
 #else // defined WITH_MAGICK
-#error "DGtal has not been built with imagemagick support. Consider adding -DWITH_MAGICK=true when building the project with cmake."
+#  error "DGtal has not been built with imagemagick support. Consider adding -DWITH_MAGICK=true when building the project with cmake."
 #endif // defined WITH_MAGICK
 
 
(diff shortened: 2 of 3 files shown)


JacquesOlivierLachaud:
Thanks Roland !

copyme:
My five cents, should not this problem be reported to the ImageMagick team?

copyme:
I mean, these warnings should be fixed, etc.

dcoeurjo:
I also have these lines on my mac. Probably generated during the configuration...

dcoeurjo:
thx @rolanddenis for the fix:)

dcoeurjo:
but travis is not happy with it ;)

rolanddenis:
Ok, this fix doesn't work...

By the way, I confused GraphicsMagick with ImageMagick... Travis use the first one while I checked the second one

dcoeurjo:
this is an issue, both implement the same Magick++ API but I only have imagemagick on my Mac too.
We should keep the vanilla imagemagick in travis

rolanddenis:
Are you sure you don't have both installed and CMake found GraphicsMagick first ? (In fact, the `FindMagick.cmake` script is unable to found ImageMagick on Ubuntu 16.04)

Anyway, I found the erroneous lines in the GraphicsMagick source and it comes from https://sourceforge.net/p/graphicsmagick/code/ci/056042125005b891a94c9834eae0623de20be2b5/
committed at Septembre 30th. They missed the `diagnostic push` :smirk:

dcoeurjo:
maybe but we should cleanup the travis script `brew install qt5 graphicsmagick fftw eigen
` to only consider imagemagick to avoid any confusion

dcoeurjo:
(and report an issue if `imagemagick` cannot be found  on ubuntu)

rolanddenis:
In fact, the cmake script is designed for GraphicsMagick, even the error message is about the dependency on GraphicsMagick. I will open an issue...

dcoeurjo:
well.. apparently, we recommend graphicsmagick at the build step ;) https://dgtal.org/doc/nightly/moduleBuildDGtal.html

rolanddenis:
:sweat_smile:

dcoeurjo:
we need to decide between the two

rolanddenis:
Does we ? If the API is the same, we should make CMake script compatible with both library and make Travis use ImageMagick instead of GraphicsMagick for the warning check job (at least until the warning is fixed).

dcoeurjo:
It would be easier to follow/report issues to one project. Clearly we have an incompatibility here between the twos.
Inconsistencies may happen in the future. I think it'd be easier to decide and drop or do not support the other one

rolanddenis:
Officially, DGtal is only compatible with GraphicsMagick, isn't it? It was only me that didn't know the difference between this library and ImageMagick :sweat_smile: 

For the record, I have submitted an issue on GraphicsMagick bug tracker :octocat:

copyme:
@rolanddenis you were not the only one here... ;-)

dcoeurjo:
historically, graphicsmagick is a fork of imagemagick with the aim to keep the same API (Magick++)

dcoeurjo:
in DGtal, we don't care a lot since we are using very basic stuff from the lib (that may have diverged, I don't know). 
But the problem is that we cannot check the two packages at the same time (and report to both upstream if we see an error). So my point would be to clarify and make a choice

dcoeurjo:
honestly I don't care but https://trends.google.com/trends/explore?q=imagemagick,graphicsmagick :D

dcoeurjo:
even if GM seems to have better perfs..

rolanddenis:
Yeah, it is quite ugly! I may add some comments to prevent a heart attack if we fall back on these lines in a while... 😆

rolanddenis:
The (temporary) fix is ready! Should I update the `ChangeLog.md`?

JacquesOlivierLachaud:
Yes. I think so. @dcoeurjo ?

dcoeurjo:
please, do

rolanddenis:
Done :octocat:

dcoeurjo:
thanks a lot.. merging.
@JacquesOlivierLachaud could you please update your branch ?

JacquesOlivierLachaud:
Done

rolanddenis:
The issue will be fixed in the next version of GraphicsMagick, see https://sourceforge.net/p/graphicsmagick/code/ci/d4a5e3d169575eda2cadfa88ea63b7f9f174c640/

Issue #1381, opened 2019-01-15
Compilation with Visual Studio


BorisMansencal:
Compilation of DGtal head fails on Windows 10 (1809) with Visual Studio Community 17 (15.9.5).

PR #1380 makes the code compile with Visual Studio (15.9.5).
It makes also the following tests pass :
testLongVol
testPNMReader
testVolReader
testGenericReader
testPNMRawWriter
testGenericWriter

In Release mode, the following 4 tests still fail:
testCloneAndAliases
testClone2
testVoxelComplex
testNeighborhoodConfigurations

In Debug mode, the code does not compile.
I get an error on tutorial-examples/shortcuts-geometry.cpp
error C1128 "number of sections exceeded object file format limit: compile with /bigobj".
The file has to be split or the CMakeLists.txt has to be changed to compile with /bigobj on Windows with Visual.

If I comment out some code in shortcults-geometry.cpp to make it compile, everything else compiles correctly.
However, 11 tests still fail:
testCloneAndAliases
testClone2
testContourHelper
testCombinFSS
testFMM
testChamferDT
testChamferVoro
testCubicalComplex
testVoxelComplex
testNeighborhoodConfigurations
testParDirCollapse

Visual Studio (15.9.5) seems to check more things regarding the STL use.
The compilation outputs some frightening error messages:
- "can't dereference out of range vector iterator"
- "cannot dereference string iterator because it is out of range"
- "map/set iterators incompatible"
- "list iterators incompatible"
...

JacquesOlivierLachaud:
Do you have more information on why tests are not passing ? I do not have VS to check these tests.
For testClone2, I had a version of gcc (5.4) where it fails while it works with clang (3.8), so I might have found the issue (upcoming FixClone), It is a related to a bad implicit conversion.
Can you provide error logs or try to tell us where you think is a the problem.

BorisMansencal:
I will try to provide more details.

BorisMansencal:
Actually, I think you can replicate at least some of the errors by compiling wih gcc with the flag -D_GLIBCXX_DEBUG
(I tested on Ubuntu 16.04 by adding this flag to CMAKE_CXX_FLAGS )

I will try to detail what I have found in the following messages.
(I am not an expert in DGtal code and concepts, so bear with me...)

BorisMansencal:
**testFMM** fail is due to the use of an invalid iterator at src/DGtal/geometry/volumes/distance/FMM.ih:416
[...]

PR #1382, opened 2019-01-16
Update CowPtr and Clone to fix testClone2

### LINKED ISSUE #1203: testClone2 failure with g++ 5.4.0 (Ubuntu 16.04)
`testClone2` compiles fine but fails at runtime (see https://gist.github.com/rolanddenis/842ac840e76c6d70a75c787ce14d0cfa ).

Everything works fine with g++ 5.2.1 (Ubuntu 15.10).

JacquesOlivierLachaud:
# PR Description

This PR fixes an issue in testClone2.cpp (and more generally in the Clone / CowPtr and CountedPtr mecanism) so that it compiles correctly both with gcc and clang (and hopefully Visual Studio).
GCC was unable to solve a construction of a CountedPtr<T> member from a Clone<T> parameter in an efficient way.

Related to #1381. Fix #1203.

# Checklist

- [x] Unit-test of your feature with testClone2.cpp

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -34,7 +34,7 @@
 - *Base*
   - Fixing wrong members in PredicateCombiner (David Coeurjolly,
     [#1321](https://github.com/DGtal-team/DGtal/pull/1321))
-  - Fix testClone2.cpp and efficiency issue in Clone/CountedPtr mecanism (Jacques-Olivier Lachaud,
+  - Fix testClone2.cpp and efficiency issue in Clone/CountedPtr mechanism (Jacques-Olivier Lachaud,
     [#1382](https://github.com/DGtal-team/DGtal/pull/1382)). Fix issue
     [#1203](https://github.com/DGtal-team/DGtal/issues/1203))
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -34,6 +34,8 @@
 - *Base*
   - Fixing wrong members in PredicateCombiner (David Coeurjolly,
     [#1321](https://github.com/DGtal-team/DGtal/pull/1321))
+  - Fix testClone2.cpp and efficiency issue in Clone/CountedPtr mecanism (Jacques-Olivier Lachaud,
+    [#1382](https://github.com/DGtal-team/DGtal/pull/1382))
 
 - *Shapes*
   - Add two new star shapes: Astroid and Lemniscate
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -35,7 +35,8 @@
   - Fixing wrong members in PredicateCombiner (David Coeurjolly,
     [#1321](https://github.com/DGtal-team/DGtal/pull/1321))
   - Fix testClone2.cpp and efficiency issue in Clone/CountedPtr mecanism (Jacques-Olivier Lachaud,
-    [#1382](https://github.com/DGtal-team/DGtal/pull/1382))
+    [#1382](https://github.com/DGtal-team/DGtal/pull/1382)). Fix issue
+    [#1203](https://github.com/DGtal-team/DGtal/issues/1203))
 
 - *Shapes*
   - Add two new star shapes: Astroid and Lemniscate
--- a/src/DGtal/base/Clone.h
+++ b/src/DGtal/base/Clone.h
@@ -272,7 +272,30 @@ namespace DGtal
     enum Parameter { CONST_LEFT_VALUE_REF, LEFT_VALUE_REF, PTR, CONST_PTR,
 		     COW_PTR, COUNTED_PTR, RIGHT_VALUE_REF, COUNTED_PTR_OR_PTR,
 		     COUNTED_CONST_PTR_OR_CONST_PTR };
-
+    /**
+       Debug method for displaying what's happening when using Clone mecanism
+       @param method the name of the conversion method
+       @param p the type of parameter
+    */
+    void display( const std::string& method, Parameter p ) const
+    {
+      std::cout << "[Clone<T>::" << method << " param=";
+      std::string sp;
+      switch (p) {
+      case CONST_LEFT_VALUE_REF: sp = "CONST_LEFT_VALUE_REF"; break;
+      case LEFT_VALUE_REF:       sp = "LEFT_VALUE_REF"; break;
+      case PTR:                  sp = "PTR"; break;
+      case CONST_PTR:            sp = "CONST_PTR"; break;
+      case COW_PTR:              sp = "COW_PTR"; break;
+      case COUNTED_PTR:          sp = "COUNTED_PTR"; break;
+      case RIGHT_VALUE_REF:      sp = "RIGHT_VALUE_REF"; break;
+      case COUNTED_PTR_OR_PTR:   sp = "COUNTED_PTR_OR_PTR"; break;
+      case COUNTED_CONST_PTR_OR_CONST_PTR: sp = "COUNTED_CONST_PTR_OR_CONST_PTR"; break;
+      default:                   sp = "UNKNOWN"; break;
+      }
+      std::cout << sp << "]" << std::endl;
+        
+    }
     /// Internal class that is used for a late deletion of an acquired pointer.
     struct TempPtr {
       /**
@@ -360,6 +383,7 @@ namespace DGtal
     */
     inline operator T() const
     {
+      // display( "operator T() const", myParam );
       switch( myParam ) {
       case CONST_LEFT_VALUE_REF:
 	return T( * static_cast< const T* >( myPtr ) );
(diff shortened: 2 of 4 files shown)


rolanddenis:
Also related to #1203

ChangeLog.md
```diff
@@ -34,6 +34,9 @@
 - *Base*
   - Fixing wrong members in PredicateCombiner (David Coeurjolly,
     [#1321](https://github.com/DGtal-team/DGtal/pull/1321))
+  - Fix testClone2.cpp and efficiency issue in Clone/CountedPtr mecanism (Jacques-Olivier Lachaud,
```
dcoeurjo:
```suggestion
  - Fix testClone2.cpp and efficiency issue in Clone/CountedPtr mechanism (Jacques-Olivier Lachaud,
```

src/DGtal/base/Clone.h
```diff
@@ -272,7 +272,30 @@ namespace DGtal
     enum Parameter { CONST_LEFT_VALUE_REF, LEFT_VALUE_REF, PTR, CONST_PTR,
 		     COW_PTR, COUNTED_PTR, RIGHT_VALUE_REF, COUNTED_PTR_OR_PTR,
 		     COUNTED_CONST_PTR_OR_CONST_PTR };
-
+    /**
+       Debug method for displaying what's happening when using Clone mecanism
```
dcoeurjo:
```suggestion
       Debug method for displaying what's happening when using Clone mechanism
```

src/DGtal/base/CowPtr.h
```diff
@@ -75,9 +75,13 @@ namespace DGtal
     // no need for ~CowPtr - the CountedPtr takes care of everything.
     CowPtr(const CowPtr& r) noexcept             : myPtr(r.myPtr) {}
     /**
-     * @todo JOL: check this.
+     * Builds a copy-on-write pointer from a counted pointer. Requires
+     * an extra dummy parameter in order to solve ambiguities when
+     * casting from Clone<T> to CowPtr<T>.
```
rolanddenis: Don't you mean from `CounterPtr` to `CowPtr` ? (but I didn't find the ambiguities you talk about)

JacquesOlivierLachaud: Before there was a constructor CowPtr<T>::CountedPtr<t>( p ). And a conversion operator Clone<T>::operator CountedPtr<T>(). So I could not write an operator Clone<T>::operator CowPtr<T>() because the compiler was complaining about ambiguities when building a CowPtr<T> from a Clone<T>.
By adding an extra parameter in the constructor CowPtr<T>::CountedPtr<t>( p, dummy ), there is no more an ambiguity and I was able to write the two conversion operator in Clone<T>.
Should I say more in the doc ?

tests/base/testClone2.cpp
```diff
@@ -948,7 +948,7 @@ int main()
 {
   bool ok = true
     && testCloneCases() 
-    && testCloneTimings()
+    //    && testCloneTimings()
```
dcoeurjo: cleanup ?

src/DGtal/base/CowPtr.h
rolanddenis: Ok I see. I have to check that a little bit more since I found it strange that we introduce such an ambiguity (but I concede that it may be necessary). Does both way (`Clone -> CountedPtr -> CowPtr` and  `Clone -> CowPtr`) lead to the same result?

JacquesOlivierLachaud:
Travis CI errors seem not to be related to this PR. Was there a recent PR modifying the travis script ?

rolanddenis:
I appends sometimes, it seems to be temporary issues with Travis... I've restarted the jobs ;)

tests/base/testClone2.cpp
JacquesOlivierLachaud: done

JacquesOlivierLachaud:
Looks ready now :-)

src/DGtal/base/CowPtr.h
rolanddenis: Ok, thanks for the additional doc!

src/DGtal/base/Clone.h
```diff
@@ -398,14 +423,43 @@ namespace DGtal
...
         return CowPtr<T>( 0 );
       }
     }
 
+    /**
+       Cast operator to a smart pointer on T instance. The object is duplicated or not
+       depending on the type of input parameter.
+
+      - const T & -> CountedPtr<T>     // immediate duplication 
+      - T* -> CountedPtr<T>            // acquired
```
rolanddenis: Shouldn't be duplicated in that case?

src/DGtal/base/CowPtr.h
```diff
@@ -75,9 +75,35 @@ namespace DGtal
...
+       @note Since 1.0, there is no more the direct constructor
+       CowPtr<T>::CowPtr( CountedPtr<T> ). Indeed, it was creating an
+       ambiguity when using conversion operator in class
+       Clone<T>. More precisely it was impossible to have two
+       conversion operators: Clone<T>::operator CountedPtr<T> and
+       Clone<T>::operator CowPtr<T> since this was creating a
+       compilation ambiguity. This is why there is now a dummy bool
+       parameter in this constructor. And there are the conversion
+       operator Clone<T>::operator CountedPtr<T> and
+       Clone<T>::operator CowPtr<T>.
```
rolanddenis: Redundant sentence?

tests/base/testClone2.cpp
```diff
@@ -241,7 +241,7 @@ struct CloneToValueMember {
 
 struct CloneToCountedMember { // requires explicit duplication
   inline CloneToCountedMember( Clone<DummyTbl> a1 ) // : myDummyTbl( a1 ) {} does not compile
-    : myDummyTbl( new DummyTbl( a1 ) ) {}
+    : myDummyTbl( a1 ) {} // : myDummyTbl( new DummyTbl( a1 ) ) {}
```
rolanddenis:
```suggestion
    : myDummyTbl( a1 ) {}
```

src/DGtal/base/Clone.h
JacquesOlivierLachaud: Every time a pointer is passed as parameter, it is considered as dyn. alloc. and handed as parameter in order to be acquired by the object (cf doc). However, I do agree that perhaps conversion to CountedPtr<T> could be forbidden from a Clone<T>. In the doc, I only talk about a pointer T* acquired by a CowPtr<T> when the user use Clone<T>. What do you think ?

src/DGtal/base/CowPtr.h
JacquesOlivierLachaud: I have rephrased it.

src/DGtal/base/CowPtr.h
```diff
@@ -75,9 +75,35 @@ namespace DGtal
...
+       @param r any counted pointer
+       
+       @note Since 1.0, there is no more the direct constructor
+       CowPtr<T>::CowPtr( CountedPtr<T> ). Indeed, it was creating an
+       ambiguity when using conversion operator in class
+       Clone<T>. More precisely it was impossible to have two
+       conversion operators: Clone<T>::operator CountedPtr<T> and
+       Clone<T>::operator CowPtr<T> since this was creating a
+       compilation ambiguity. This is why there is now a dummy bool
+       parameter in this constructor while both conversion operator
```
rolanddenis:
```suggestion
       parameter in this constructor while both conversion operators
```

src/DGtal/base/CowPtr.h
rolanddenis: Thx, just a typo at operator (see other comment)

src/DGtal/base/Clone.h
rolanddenis: Ok for the acquirement of the pointer, thanks to the doc.

About the conversion to a `CountedPtr` when the `Clone` is initialized from a pointer, the problem is that if the created `CountedPtr` has a shorter lifetime than the `Clone`, it will lead to an invalid pointer (freed by `CountedPtr`). Is that a valid usage of `Clone` ?

While searching a realistic example to the case above, may we have to pass a `Clone` to a `(Const)Alias`?

rolanddenis: But maybe `Clone` is supposed to be transferred into another holder before being used?

JacquesOlivierLachaud: Arg. I dunno. It should work in most instances but I am not sure.
Here I think it cannot work (or risk of ambiguity) since I have no constructor ConstAlias<T> from Clone<T>.

rolanddenis: Ok. Anyway, It's not the time to discuss on this particular point, so we will check that later ;)

dcoeurjo:
All good for me too. Thx @JacquesOlivierLachaud and @rolanddenis !

Merging (can you please close the associated Issues ?)

PR #1383, opened 2019-01-16
Fix Travis (documentation deploy & DGtalTools)

rolanddenis:
# PR Description
Update and fix of Travis script:
- using classical environment variables (`PATH`, `CPATH` and `CMAKE_PREFIX_PATH`) instead of keeping path of each dependency.
- calling script instead or sourcing it when possible.
- chaining script lines so that it terminates if one step fails (but we loose the separate timings).
- re-enabling DGtalTools compilation.
- fixing documentation deploy.

# Checklist

- [x] https://github.com/DGtal-team/DGtalTools/pull/345 is merged in DGtalTools
- [x] Removing temporary fix that ignore DGtalTools compilation failure

### DIFF
--- a/.travis/build_all.sh
+++ b/.travis/build_all.sh
@@ -1,44 +1,42 @@
 #!/bin/bash
 $SCRIPT_BEGIN
 
-
-export CONFIG="Debug,Magick,GMP,ITK,FFTW3,Debug,Cairo,QGLviewer,HDF5,EIGEN"
-
 # OS dependent deps
-source "$SRC_DIR/.travis/install_eigen.sh"
-echo $EIGEN_ROOT
+"$SRC_DIR/.travis/install_eigen.sh"
 
-export BTYPE="$BTYPE -DBUILD_EXAMPLES=false -DBUILD_TESTING=true"
-export BTYPE="$BTYPE -DCMAKE_BUILD_TYPE=Debug -DWITH_MAGICK=true -DWITH_GMP=true\
-                     -DWITH_FFTW3=true -DWARNING_AS_ERROR=ON -DCMAKE_BUILD_TYPE=Debug \
-                     -DWITH_HDF5=true -DWITH_CAIRO=true -DWITH_QGLVIEWER=true -DWITH_EIGEN=true\
-                     -DWARNING_AS_ERROR=OFF -DEIGEN3_INCLUDE_DIR='$EIGEN_ROOT/include/eigen3'"
+# Build options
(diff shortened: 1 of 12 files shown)


src/DGtal/base/Common.h
```diff
@@ -111,7 +111,7 @@
 #include "DGtal/base/Exceptions.h"
 
 //////////////////////////////////////////////////////////////////////////////
-namespace DGtal
+namespaced DGtal
```
JacquesOlivierLachaud: What is the meaning of "namespaced" in C++ ?

rolanddenis: None, it is meant to trigger a compilation error to test Travis behaviour :p

JacquesOlivierLachaud: ok :)

.travis/build_all.sh
```diff
@@ -1,44 +1,43 @@
...
 #############################
-echo "Preparing the build..."
+echo "Building..."
 
 #Sequential DEC examples, this would also build library
 #make examplePropagation
 ninja testDiscreteExteriorCalculusExtended
 #make exampleDiscreteExteriorCalculusChladni
 ninja
+ninja install
```
dcoeurjo: why do you need the install?

.travis/build_dgtal.sh
```diff
@@ -35,8 +31,8 @@ cmake "$SRC_DIR" $BTYPE -G Ninja
 #############################
 #     make all
 #############################
-echo "Preparing the build..."
-#cd "$BUILD_DIR"
-#Sequential DEC examples, this would also build library
+echo "Building..."
 ninja
+ninja install
```
dcoeurjo: same here

.travis/build_all.sh
rolanddenis: Thus (following the comment below), it is not necessary here since we don't use DGtal after this compilation. I only add this to copy the `build_dgtal.sh` script ... should I remove it ?

dcoeurjo: If you don't mind... This would just clean it up

.travis/build_dgtal.sh
dcoeurjo: 👌  thx. it makes, sense.

.travis/build_all.sh
rolanddenis: 😁

dcoeurjo:
Thanks @rolanddenis , merging

PR #1345, opened 2018-09-10
Fixing PointVector implicit conversions

rolanddenis:
# PR Description

Fixing conversions between `PointVector` with different Euclidean Ring type and storage (see #1341).
Automatic Euclidean Ring return type for compatible operations.

Note that all external operators (`+`, `-`, `<`, ...) and functions (`dotProduct`, `crossProduct`, ...) that are related to `PointVector` are declared as friend of this class. It is not necessary since those operators and functions never access to protected/private members but it allows to have them documented on the same page as `PointVector` in a "Friends" section.

# Checklist

- [ ] Unit-test of your feature with [Catch](http://dgtal.org/doc/stable/moduleCatch.html).
- [ ] Doxygen documentation of the code completed (classes, methods, types, members...)
- [ ] Documentation module page added or updated.
- [ ] New entry in the [ChangeLog.md](https://github.com/DGtal-team/DGtal/blob/master/ChangeLog.md) added.
- [ ] No warning raised in Debug ```cmake``` mode (otherwise, Travis C.I. will fail).
- [ ] All continuous integration tests pass (Travis & appveyor)
- [ ] Clean commit history

# PointVector checklist
- [x] Assignment operator
- [x] Unary & binary functor conversion constructors
- [x] Arithmetic binary operators
- [x] Arithmetic unary operators with assignment (`+=`, `-=`,...)
- [x] Comparison operators (using lexicographical order)
- [x] Other binary functions on `PointVector` (`dot`, `crossProduct`, `inf`,...) . Also external versions.
- [x] Adding comparison tests
- [x] `partialCopy` and `partialEqual`
- [x] New functors to ease the choice of rounding mode without specifying type (like with `std::round`)
- [x] ~~`crossProduct` of 2D vectors returns a scalar. This is a critical change that may lead to compilation errors (e.g. https://github.com/rolanddenis/DGtal/blob/fix_pointvector/src/DGtal/shapes/Mesh.ih#L464)~~

# Conversion fixes
### Why is there a conversion issue ?
When a `PointVector<dim,T>` is converted into another `PointVector<dim,U>` in a implicit way (e.g. when passing the first type to a function accepting the second type) and if converting from `T` to `U` is ambiguous (e.g. from `double` to `int`), then this conversion will fail.

### Which conversions are valid ?
[...]

PR #1388, opened 2019-01-27
LpMetric in R^d

dcoeurjo:
# PR Description

This PR introduces a generic LpMetric class on double (to fix Point<->RealPoint implicit casts when considering the distance between two (integer) Points).

# Checklist

### DIFF
--- a/src/DGtal/geometry/doc/moduleMetrics.dox
+++ b/src/DGtal/geometry/doc/moduleMetrics.dox
@@ -69,7 +69,7 @@ example,
 on C++ "double").
 - @f$ l_p@f$ metrics (ExactPredicateLpSeparableMetric or
   InexactPredicateLpSeparableMetric) are models of CMetricSpace on @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$ for
-  @f$p>1@f$ with specific separabilty methods.
+  @f$p>1@f$ with specific separability methods.
 - Chamfer norms with axis symmetric unit ball (see experimental::ChamferNorm2D in dimension 2).
 - DigitalMetricAdapter which adapts any metric (model of CMetricSpace)
   to digital space using the ceil function. This induces a digital
--- a/src/DGtal/geometry/doc/moduleMetrics.dox
+++ b/src/DGtal/geometry/doc/moduleMetrics.dox
@@ -62,12 +62,14 @@ In the following, we restrict our attention to metric spaces even if some of the
 In DGtal, the concept of CMetricSpace is associated with metric space
 definition. A first refinement of this concept is the concept of
 CDigitalMetricSpace whose models implement distance functions on
-integer numbers (i.e. @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$)). For
+integer numbers (i.e. @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$). For
 example,
 
+- LpMetric are models of CMetricSpace on @f$(\mathbb{R}^n, \mathbb{R}, d)@f$ spaces (i.e. all computations are performed
+on c++ "double").
 - @f$ l_p@f$ metrics (ExactPredicateLpSeparableMetric or
-  InexactPredicateLpSeparableMetric) are models of CMetricSpace for
-  @f$p>1@f$.
+  InexactPredicateLpSeparableMetric) are models of CMetricSpace on @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$ for
+  @f$p>1@f$ with specific separabilty methods.
 - Chamfer norms with axis symmetric unit ball (see experimental::ChamferNorm2D in dimension 2).
 - DigitalMetricAdapter which adapts any metric (model of CMetricSpace)
   to digital space using the ceil function. This induces a digital
--- a/src/DGtal/geometry/doc/moduleMetrics.dox
+++ b/src/DGtal/geometry/doc/moduleMetrics.dox
@@ -66,7 +66,7 @@ integer numbers (i.e. @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$). For
 example,
 
 - LpMetric are models of CMetricSpace on @f$(\mathbb{R}^n, \mathbb{R}, d)@f$ spaces (i.e. all computations are performed
-on c++ "double").
+on C++ "double").
 - @f$ l_p@f$ metrics (ExactPredicateLpSeparableMetric or
   InexactPredicateLpSeparableMetric) are models of CMetricSpace on @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$ for
   @f$p>1@f$ with specific separabilty methods.
--- a/src/DGtal/geometry/volumes/distance/LpMetric.h
+++ b/src/DGtal/geometry/volumes/distance/LpMetric.h
@@ -183,7 +183,7 @@ namespace DGtal
       if (dfirst < dsecond)
         return ClosestFIRST;
       else
-        if (dfirst>dsecond)
+        if (dfirst > dsecond)
           return ClosestSECOND;
       
       return ClosestBOTH;
--- a/src/DGtal/geometry/volumes/distance/LpMetric.h
+++ b/src/DGtal/geometry/volumes/distance/LpMetric.h
@@ -23,7 +23,6 @@
  *
  * @date 2019/27/01
  *
- * Header file for module LpMetric.cpp
  *
  * This file is part of the DGtal library.
  */
--- a/src/DGtal/geometry/volumes/distance/LpMetric.h
+++ b/src/DGtal/geometry/volumes/distance/LpMetric.h
@@ -97,7 +97,7 @@ namespace DGtal
       /**
      * Constructor.
      *
-     * @param anExponent the exponent (p) of the lp metric.
+     * @param anExponent the exponent (@a p) of the lp metric.
      */
     LpMetric( const double anExponent): myExponent(anExponent)
     {}
--- a/src/DGtal/geometry/volumes/distance/LpMetric.h
+++ b/src/DGtal/geometry/volumes/distance/LpMetric.h
@@ -216,7 +216,11 @@ namespace DGtal
    */
   template <typename T>
   std::ostream&
-  operator<< ( std::ostream & out, const LpMetric<T> & object );
+  operator<< ( std::ostream & out, const LpMetric<T> & object )
+  {
+    object.selfDisplay( out );
+    return out;
+  }
 
 } // namespace DGtal
 
--- a/src/DGtal/geometry/volumes/distance/LpMetric.h
+++ b/src/DGtal/geometry/volumes/distance/LpMetric.h
@@ -88,13 +88,6 @@ namespace DGtal
     typedef double RawValue;
     
     /**
-     * Default Constructor (p=2.0)
-     *
-     */
-    LpMetric(): myExponent(2.0) {}
-    
-
-      /**
      * Constructor.
      *
      * @param anExponent the exponent (@a p) of the lp metric.
@@ -158,8 +151,8 @@ namespace DGtal
     {
       RawValue tmp=0.0;
       for(auto i = 0; i < aP.size(); ++i)
-        tmp += std::pow(NumberTraits<typename Point::Coordinate>::castToDouble(std::abs(aP[i] - aQ[i])),
-                        myExponent);
+        tmp += static_cast<RawValue>(std::pow(NumberTraits<typename Point::Coordinate>::castToDouble(std::abs(aP[i] - aQ[i])),
+                                              myExponent));
       return tmp;
     }
 
--- a/src/DGtal/geometry/volumes/distance/LpMetric.h
+++ b/src/DGtal/geometry/volumes/distance/LpMetric.h
@@ -55,7 +55,7 @@ namespace DGtal
  * Description of template class 'LpMetric' <p>
  * \brief Aim: implements  l_p metrics.
  *
- * Given a parameter p, the class implement classical l_p
+ * Given a parameter p, the class implements classical l_p
  * metric as a model of CMetricSpace. Hence, given two points
  * @f$ x=(x_0...x_{n-1})@f$, @f$ y=(y_0...y_{n-1})@f$, we define a metric as:
  *
@@ -156,7 +156,7 @@ namespace DGtal
      */
     RawValue rawDistance(const Point & aP, const Point &aQ) const
     {
-      double tmp=0.0;
+      RawValue tmp=0.0;
       for(auto i = 0; i < aP.size(); ++i)
         tmp += std::pow(NumberTraits<typename Point::Coordinate>::castToDouble(std::abs(aP[i] - aQ[i])),
                         myExponent);
--- a/src/DGtal/geometry/volumes/distance/LpMetric.h
+++ b/src/DGtal/geometry/volumes/distance/LpMetric.h
@@ -150,7 +150,7 @@ namespace DGtal
     RawValue rawDistance(const Point & aP, const Point &aQ) const
     {
       RawValue tmp=0.0;
-      for(auto i = 0; i < aP.size(); ++i)
+      for(typename Point::Dimension i = 0; i < aP.size(); ++i)
         tmp += static_cast<RawValue>(std::pow(NumberTraits<typename Point::Coordinate>::castToDouble(std::abs(aP[i] - aQ[i])),
                                               myExponent));
       return tmp;
--- a/src/DGtal/geometry/volumes/distance/LpMetric.h
+++ b/src/DGtal/geometry/volumes/distance/LpMetric.h
@@ -1,0 +1,237 @@
+/**
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Lesser General Public License as
+ *  published by the Free Software Foundation, either version 3 of the
+ *  License, or  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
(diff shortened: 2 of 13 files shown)


dcoeurjo:
Should fix #1345 issues

dcoeurjo:
@rolanddenis : for `dvcm-2d` and `shortcut` I need to investigate a bit more.  E.g. `dvcm` needs a Separable metric and I don't see where the issue is

src/DGtal/geometry/volumes/distance/LpMetric.h
```diff
@@ -0,0 +1,237 @@
...
+     * @param aP a first point.
+     * @param aQ a second point.
+     *
+     * @return the distance between aP and aQ.
+     */
+    RawValue rawDistance(const Point & aP, const Point &aQ) const
+    {
+      double tmp=0.0;
+      for(auto i = 0; i < aP.size(); ++i)
+        tmp += std::pow(NumberTraits<typename Point::Coordinate>::castToDouble(std::abs(aP[i] - aQ[i])),
```
rolanddenis: Same (like in `ExactPredicateLpSeparableMetric`actually)

src/DGtal/geometry/volumes/distance/LpMetric.h
```diff
@@ -0,0 +1,237 @@
...
+    
+    ///Type for points (RealPoint for this class)
+    typedef typename Space::RealPoint Point;
+    ///Type for distance values
+    typedef double Value;
+    ///Type for raw distance values
+    typedef double RawValue;
+    
+    /**
+     * Default Constructor (p=2.0)
```
rolanddenis: I'm not completely comfortable with having a default constructor. When looking at code like
```C++
LpMetric<Space> metric;
```
I found that strange to not specify the `p`  and will have to look at the documentation to check the default behavior. I prefer to remove the default constructor and to be forced to always specify the power:
```C++
LpMetric<Space> metric(2);
```
Don't you?

tests/geometry/volumes/distance/testLpMetric.cpp
```diff
@@ -0,0 +1,67 @@
...
+  LpMetric<Z2i::Space> l55_2D(5.5);
+  Z2i::Space::RealPoint a(0,0), b(1.0,1.0);
+  Z3i::Space::RealPoint aa(0,0,0), bb(1.0,1.0,1.0);
+  
+  SECTION("Testing LpMetric distance values")
+    {
+      REQUIRE( l2_2D.rawDistance(a,b) == Approx(2.0) );
+      REQUIRE( l2_3D.rawDistance(aa,bb) == Approx(3.0) );
+      REQUIRE( l2_2D(a,b) == Approx(std::sqrt(2.0)) );
+      REQUIRE( l55_2D(a,b) == Approx(1.1343125));
```
rolanddenis: We should also test the `closest` feature (maybe the `operator()` also)

rolanddenis:
So, the only difference with `InexactPredicateLpSeparableMetric` is the `Point` type?

src/DGtal/geometry/volumes/distance/LpMetric.h
```diff
@@ -0,0 +1,237 @@
...
+#pragma once
+
+/**
+ * @file LpMetric.h
+ * @author David Coeurjolly (\c david.coeurjolly@liris.cnrs.fr )
+ * Laboratoire d'InfoRmatique en Image et Systèmes d'information - LIRIS (CNRS, UMR 5205), CNRS, France
+ *
+ * @date 2019/27/01
+ *
+ * Header file for module LpMetric.cpp
```
dcoeurjo:
```suggestion
```

src/DGtal/geometry/doc/moduleMetrics.dox
```diff
@@ -62,12 +62,14 @@ In the following, we restrict our attention to metric spaces even if some of the
 In DGtal, the concept of CMetricSpace is associated with metric space
 definition. A first refinement of this concept is the concept of
 CDigitalMetricSpace whose models implement distance functions on
-integer numbers (i.e. @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$)). For
+integer numbers (i.e. @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$). For
 example,
 
+- LpMetric are models of CMetricSpace on @f$(\mathbb{R}^n, \mathbb{R}, d)@f$ spaces (i.e. all computations are performed
+on c++ "double").
```
dcoeurjo:
```suggestion
on C++ "double").
```

src/DGtal/geometry/volumes/distance/LpMetric.h
```diff
@@ -0,0 +1,237 @@
...
+namespace DGtal
+{
+
+/////////////////////////////////////////////////////////////////////////////
+// template class LpMetric
+/**
+ * Description of template class 'LpMetric' <p>
+ * \brief Aim: implements  l_p metrics.
+ *
+ * Given a parameter p, the class implement classical l_p
```
dcoeurjo:
```suggestion
 * Given a parameter p, the class implements classical l_p
```

src/DGtal/geometry/volumes/distance/LpMetric.h
dcoeurjo:
```suggestion
```

src/DGtal/geometry/doc/moduleMetrics.dox
```diff
@@ -62,12 +62,14 @@ In the following, we restrict our attention to metric spaces even if some of the
...
 - @f$ l_p@f$ metrics (ExactPredicateLpSeparableMetric or
-  InexactPredicateLpSeparableMetric) are models of CMetricSpace for
-  @f$p>1@f$.
+  InexactPredicateLpSeparableMetric) are models of CMetricSpace on @f$(\mathbb{Z}^n, \mathbb{Z}, d)@f$ for
+  @f$p>1@f$ with specific separabilty methods.
```
dcoeurjo:
```suggestion
  @f$p>1@f$ with specific separability methods.
```

src/DGtal/geometry/volumes/distance/LpMetric.h
```diff
@@ -0,0 +1,237 @@
...
+     * Compute the raw distance between @a aP and @a aQ.
+     *
+     * @param aP a first point.
+     * @param aQ a second point.
+     *
+     * @return the distance between aP and aQ.
+     */
+    RawValue rawDistance(const Point & aP, const Point &aQ) const
+    {
+      double tmp=0.0;
```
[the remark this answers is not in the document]
dcoeurjo: Agree;)

dcoeurjo:
```suggestion
      RawValue tmp=0.0;
```

src/DGtal/geometry/volumes/distance/LpMetric.h
dcoeurjo: what do you mean ? a static_cast to `RawValue` after the `castToDouble` ?

tests/geometry/volumes/distance/testLpMetric.cpp
dcoeurjo: yep

dcoeurjo:
> So, the only difference with `InexactPredicateLpSeparableMetric` is the `Point` type?

And the fact that `InexactPredicateLpSeparableMetric` is a model of CSeparableMetric

src/DGtal/geometry/volumes/distance/LpMetric.h
dcoeurjo: Agree, this would be clearer

src/DGtal/geometry/volumes/distance/LpMetric.h
```diff
@@ -0,0 +1,236 @@
...
+     * Default Constructor (p=2.0)
+     *
+     */
+    LpMetric(): myExponent(2.0) {}
+    
+
+      /**
+     * Constructor.
+     *
+     * @param anExponent the exponent (p) of the lp metric.
```
dcoeurjo:
```suggestion
     * @param anExponent the exponent (@a p) of the lp metric.
```

tests/geometry/volumes/distance/testLpMetric.cpp
dcoeurjo: done, thanks

src/DGtal/geometry/volumes/distance/LpMetric.h
dcoeurjo: NumberTraits can capture GMP numbers for instance which cannot being static_casted 

A bit strange BTW.. Point components are "double" so is RawValue here..

dcoeurjo:
n,  just a model of `CMetricSpace`

src/DGtal/geometry/volumes/distance/LpMetric.h
rolanddenis: Ok, I agree!

dcoeurjo:
thanks for the feedbacks. I've removed the default constructor

src/DGtal/geometry/volumes/distance/LpMetric.h
dcoeurjo: done

src/DGtal/geometry/volumes/distance/LpMetric.h
dcoeurjo: I've added a cast to RawValue

src/DGtal/geometry/volumes/distance/LpMetric.h
```diff
@@ -0,0 +1,237 @@
...
+
+  /**
+   * Overloads 'operator<<' for displaying objects of class 'LpMetric'.
+   * @param out the output stream where the object is written.
+   * @param object the object of class 'LpMetric' to write.
+   * @return the output stream after the writing.
+   */
+  template <typename T>
+  std::ostream&
+  operator<< ( std::ostream & out, const LpMetric<T> & object );
```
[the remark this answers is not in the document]
dcoeurjo: oops you're right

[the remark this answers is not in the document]
dcoeurjo: done

rolanddenis:
> @rolanddenis : for `dvcm-2d` and `shortcut` I need to investigate a bit more. E.g. `dvcm` needs a Separable metric and I don't see where the issue is

About that: so making `LpMetric` a `CSeparableMetric` model and using it in `dvcm-2d-curvature` should solve the issue ?

Moreover, looking at how `LocalEstimatorFromSurfelFunctorAdapter` is implemented (see https://github.com/DGtal-team/DGtal/blob/master/src/DGtal/geometry/surfaces/estimation/LocalEstimatorFromSurfelFunctorAdapter.ih#L170):
```C++
const MetricToPoint metricToPoint = std::bind( *myMetric, myEmbedder( *it ), std::placeholders::_1 );
```
with
```C++
typedef std::function< typename Metric::Value ( typename Metric::Point ) > MetricToPoint;
```
and since `CCellEmbedder` concept requires the returned value to be a `Space::RealPoint`, it implies that `LocalEstimatorFromSurfelFunctorAdapter` can only work with `LpMetric`.

Is that the wanted behaviour? (it is a real question, I'm not aware of all the usages of this class)

dcoeurjo:
> > @rolanddenis : for `dvcm-2d` and `shortcut` I need to investigate a bit more. E.g. `dvcm` needs a Separable metric and I don't see where the issue is
> 
> About that: so making `LpMetric` a `CSeparableMetric` model and using it in `dvcm-2d-curvature` should solve the issue ?

dvcm-2d needs an "exact" VoronoiMap. Can you spot me where the conversion issue is in the dvcm code ? 

I'd keep the following principles (maybe it should be clearer un the doc): 
  - LpMetric:  on R^n with `double` (model of CMetricSpace only)
  - Inexact...Separable...: on Z^n with `double`
  - Exact...Separable...: on Z^n with exact computations

Promoting LpMetric to CSeparableMetric would probably be confusing.
 

> Moreover, looking at how `LocalEstimatorFromSurfelFunctorAdapter` is implemented (see https://github.com/DGtal-team/DGtal/blob/master/src/DGtal/geometry/surfaces/estimation/LocalEstimatorFromSurfelFunctorAdapter.ih#L170):
> 
> ```c++
> const MetricToPoint metricToPoint = std::bind( *myMetric, myEmbedder( *it ), std::placeholders::_1 );
> ```
> 
> with
> 
> ```c++
> typedef std::function< typename Metric::Value ( typename Metric::Point ) > MetricToPoint;
> ```
> 
> and since `CCellEmbedder` concept requires the returned value to be a `Space::RealPoint`, it implies that `LocalEstimatorFromSurfelFunctorAdapter` can only work with `LpMetric`.
> 
> Is that the wanted behaviour? (it is a real question, I'm not aware of all the usages of this class)

I agree.. I was also thinking to force the metric in LocalEstimator.. to be a LpMetric but this would have been less generic (the only requirement is that the metric should be in R^n, and so far we have a unique model of that that is LpMetic)

rolanddenis:
> dvcm-2d needs an "exact" VoronoiMap. Can you spot me where the conversion issue is in the dvcm code ?

It comes from [`dvcm-2d-curvature.cpp:114`](https://github.com/DGtal-team/DGtal/blob/master/examples/geometry/surfaces/dvcm-2d-curvature.cpp#L114) that calls the `VCMDigitalSurfaceLocalEstimator::setParams` method:
```C++
typedef VCMDigitalSurfaceLocalEstimator<DigitalSurfaceContainer,Metric,
    KernelFunction, CurvatureFunctor> VCMCurvatureEstimator;
...
VCMCurvatureEstimator estimator( ptrSurface );
estimator.setParams( Pointels, R, r, chi, 3.0, l2, true );
```
which constructs a `VoronoiCovarianceMeasureOnDigitalSurface` (alias `VCMOnSurface`) from [`VCMDigitalSurfaceLocalEstimator::setParams:148`](https://github.com/DGtal-team/DGtal/blob/master/src/DGtal/geometry/surfaces/estimation/VCMDigitalSurfaceLocalEstimator.ih#L148):
```C++
{
  mySurfelEmbedding = surfelEmbedding;
  myVCMOnSurface = CountedConstPtrOrConstPtr<VCMOnSurface>
    ( new VCMOnSurface( mySurface, mySurfelEmbedding,
                        R, r, chi_r, t, aMetric, verbose ), true );
  myGeomFct.attach( myVCMOnSurface );
}
```
whose constructor uses a `LocalEstimatorFromSurfelFunctorAdapter` [here](https://github.com/DGtal-team/DGtal/blob/master/src/DGtal/geometry/surfaces/estimation/VoronoiCovarianceMeasureOnDigitalSurface.ih#L122):
```C++
typedef LocalEstimatorFromSurfelFunctorAdapter< DigitalSurfaceContainer, Metric, SurfelFunctor, Functor>
    NormalEstimator;
...
NormalEstimator estimator;
...
      // get rough estimation of normal
      normals.trivialNormal = estimator.eval( it );
```

rolanddenis:
> I agree.. I was also thinking to force the metric in LocalEstimator.. to be a LpMetric but this would have been less generic (the only requirement is that the metric should be in R^n, and so far we have a unique model of that that is LpMetic)

Ok, so if it is necessary, we maybe should check it in `LocalEstimatorFromSurfelFunctorAdapter` by comparing the `Value` typedef of the given metric with the `RealPoint` of the associated space (except if it is too strict) and specify it in the documentation...

dcoeurjo:
👌 thx. 
Dvcm shouldn’t use the same metric for both the Voronoi and the neighborhood exploration. Let me update that.

dcoeurjo:
I've pushed an edit in VCM (ping @JacquesOlivierLachaud ) to force the LpMetric in R^n to be used when performing the breath first propagation on surfaces.

dcoeurjo:
The question is : Should be remove the `Metric` template parameter in `LocalEstimatorFromSurfelFunctorAdapter` to force the class to use the l_2 on R^n? 
(and add the "p" value in the constructor)

src/DGtal/geometry/volumes/distance/LpMetric.h
```diff
@@ -0,0 +1,233 @@
...
+     *
+     * @param aP a first point.
+     * @param aQ a second point.
+     *
+     * @return the distance between aP and aQ.
+     */
+    RawValue rawDistance(const Point & aP, const Point &aQ) const
+    {
+      RawValue tmp=0.0;
+      for(auto i = 0; i < aP.size(); ++i)
```
rolanddenis:
```suggestion
      for(typename Point::Dimension i = 0; i < aP.size(); ++i)
```
Otherwise, there is a warning because of a signed/unsigned comparison.

dcoeurjo: 👍  thanks

dcoeurjo:
@rolanddenis ok for you ?

thx

dcoeurjo:
thanks :)

dcoeurjo:
Let's merge this one

rolanddenis:
Yep! Only a question about the metric passed to `LocalEstimatorFromSurfelFunctorAdapter` in the `VoronoiCovarianceMeasureOnDigitalSurface` constructor: it can only be a L2 metric?

dcoeurjo:
there are 2 metrics: one for the Voronoi map (must be separable) and one for the breadth first propagation (LocalEst...) qui is fixed to l2 in my edit

dcoeurjo:
We can imagine to change the Voronoi Map metric to something else for the VCM but it won't make sense to change the propagation IMHO..

rolanddenis:
Ok, my question was about the choice for the propagation, so all right :+1:

rolanddenis:
Didn't check the output of dvcm-2d-curvature before: I get a uniformly blue digitalized circle ... shouldn't we get a kind of color gradient depending on the curvature?

JacquesOlivierLachaud:
You should get something like this on an ellipse (bottom of page)
https://dgtal.org/doc/stable/moduleVCM.html
On a circle, it is correct to get a uniform blue (constant curvature), but unlikely :)

rolanddenis:
Huh yes, you're right! I found the result too simple and think there were a bug... The output seems correct for an ellipse (but not as contrasted as in the module doc but it should be due to the colormap)

PR #1395, opened 2019-02-19
Correcting Cmake files for Windows and Visual Studio, and adding d suffix for the generated library in debug mode

raphaellenain:
# PR Description

*Adding a debug version of Library (DGtald.lib)*
*Fixing CMake files for Visual Studio and Windows paths (adding quotes for paths with spaces like "C:/Program Files/DGtal")*

# Checklist


dcoeurjo:
Thanks @raphaellenain for the PR. 

Just a quick clarification: doc and dox are two specific targets in the cmake. 
The first one builds the entire doc while the second one only builds the module/package doc pages. Both are thus valid. 

(Not sure btw if the dox target is oftenly used...)

cmake/Common.cmake
```diff
@@ -36,8 +41,12 @@ message(STATUS "Checking if doxygen/dot is installed:")
 message(STATUS " ")
 set(INSTALL_DOC_PATH ${CMAKE_INSTALL_PREFIX}/share/DGtal CACHE PATH "Installation directory for DGtal documentation files.")
 INCLUDE(doxygen)
-INCLUDE(TargetDoxygenDoc OPTIONAL)
-INCLUDE(TargetDoxygenDox OPTIONAL)
+if (MSVC)
+  INCLUDE(TargetDoxygenDox OPTIONAL)
+else (MSVC)
```
dcoeurjo: Doc and dox are two (relevant) separate things. Why commenting the Doc one in MS ?

cmake/FindFFTW3.cmake
```diff
@@ -36,7 +38,9 @@ FIND_PACKAGE_HANDLE_STANDARD_ARGS(FFTW3_LONG   DEFAULT_MSG FFTW3_INCLUDE_DIR FFT
 # Adding threaded version of FFTW3
 if (FFTW3_THREADS_LIBRARIES)
   set (FFTW3_LIBRARIES ${FFTW3_LIBRARIES} ${FFTW3_THREADS_LIBRARIES})
-  set (FFTW3_DEP_LIBRARIES ${FFTW3_DEP_LIBRARIES} pthread)
+  if (NOT MSVC)
+    set (FFTW3_DEP_LIBRARIES ${FFTW3_DEP_LIBRARIES} pthread)
+  endif (NOT MSVC)
```
dcoeurjo: Is it because pthread is not available on MSW?

cmake/TargetDoxygenDox.cmake
```diff
@@ -66,7 +66,7 @@ IF (DOXYGEN_FOUND)
       STRING(REGEX REPLACE "[/]" "\\\\" HHP_FILE ${TMP} )
       # MESSAGE(SEND_ERROR "DBG  HHP_FILE=${HHP_FILE}")
       ADD_CUSTOM_TARGET(winhelp ${HTML_HELP_COMPILER} ${HHP_FILE})
-      ADD_DEPENDENCIES (winhelp doc)
+      ADD_DEPENDENCIES (winhelp dox)
```
dcoeurjo: Same comment as above.. what was the problem ?

examples/tutorial-examples/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ SET(DGTAL_TUTO_EXAMPLES_SRC
 
 
 FOREACH(FILE ${DGTAL_TUTO_EXAMPLES_SRC})
+  if(MSVC)
+    set_source_files_properties( ${FILE} PROPERTIES COMPILE_FLAGS "/bigobj" )
```
dcoeurjo: In not familiar with MS builds. Why do you need to add the `/bigbuild`?
If it is required, do we need to add this to all targets (examples, tests, dgtaltools...) ?

cmake/Common.cmake
raphaellenain: For Visual Studio, we have to choose between Doc and Dox, because the target "winhelp" is referenced in "cmake/TargetDoxygenDoc.cmake" (line 98) and in "cmake/TargetDoxygenDox.cmake" (line 68), so there is an error during the cmake process if we keep the two "doc" and "dox". I made the choice of dox, but I could have chosen doc. Maybe renaming the target "winhelp" with, for example, "winhelpx" in "cmake/TargetDoxygenDox.cmake" would work ...

cmake/FindFFTW3.cmake
raphaellenain: If we want to use pthread on Visual Studio, we have to install another dependance, while the compilation works without this library (pthread), including FFTW3 in cmake process.

raphaellenain: https://stackoverflow.com/questions/48894212/visual-studio-2017-how-to-make-include-pthread-h-work

cmake/TargetDoxygenDox.cmake
raphaellenain: the problem here is that we create a dependence named "dox" line 56, and here we want to attach "doc", which is not referenced in this file.

examples/tutorial-examples/CMakeLists.txt
raphaellenain: this flag is only needed by the example project "shortcuts-geometry" in the debug configuration. When compiling, the error is : https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1128?view=vs-2017
The fix could have been, but too much specific for my point of view  : 
```
FOREACH(FILE ${DGTAL_TUTO_EXAMPLES_SRC})
  add_executable(${FILE} ${FILE})
  target_link_libraries (${FILE} DGtal  )
ENDFOREACH(FILE)

if(MSVC)
  set_source_files_properties( shortcuts-geometry PROPERTIES COMPILE_FLAGS "/bigobj" )
endif (MSVC)
```

cmake/FindFFTW3.cmake
dcoeurjo: 👌

cmake/Common.cmake
dcoeurjo: I'd be in favor of the additional `winhelp-dox` target..

examples/tutorial-examples/CMakeLists.txt
dcoeurjo: ok.. Has this issue only been raised for this shortcut example ?

raphaellenain: Yes, this was notified by @BorisMansencal in this issue : https://github.com/DGtal-team/DGtal/issues/1381

cmake/Common.cmake
raphaellenain: Ok, I will correct this

raphaellenain:
I don't understand the error on Travis ... it's written "canceled", I don't know why ... Any idea ?

dcoeurjo:
Some admin stuff on Travis side. If you log in on Travis, you can restart the job

examples/tutorial-examples/CMakeLists.txt
dcoeurjo: 👌

cmake/Common.cmake
```diff
@@ -28,6 +28,11 @@ foreach(p LIB BIN INCLUDE DATA)
   endif()
 endforeach()
 
+# -----------------------------------------------------------------------------
+# To distinguish between debug and release lib
+# -----------------------------------------------------------------------------
+set(CMAKE_DEBUG_POSTFIX "d")
```
dcoeurjo: Would this also apply on linux/mac builds ?

dcoeurjo:
Thanks for the edits. Could you please merge the master to this branch and fix the conflict in `Changelog.md`?

cmake/Common.cmake
raphaellenain: On Linux, if we want the indicate the debug or release configuration, the "-DCMAKE_BUILD_TYPE" option has to be added : 
 - for Debug : "cmake -DCMAKE_BUILD_TYPE=Debug"
 - for Release : "cmake -DCMAKE_BUILD_TYPE=Debug" (or "cmake", Release is the default configuration)
With this new line in Common.cmake(set(CMAKE_DEBUG_POSTFIX "d")) : 
 - for Debug configuration, the generated file is libdgtald.so (on Windows : DGtald.lib)
 - for Release configuration, the generated file is libdgtal.so (on Windows : DGtal.lib)

For Mac I cannot perform any test, but I imagine it's the same thing than on Linux.

dcoeurjo: I know that but was just wondering if the "d" suffix would also be used by cmake on linux for the dynamic lib construction.

dcoeurjo: which is not standard for me

dcoeurjo: (and would imply a tweak of the FindDGtal)

dcoeurjo: What is the issue on windows if you remove this suffix ?

raphaellenain: Maybe I can add the MSVC test ? 
`if (MSVC)`
`  set(CMAKE_DEBUG_POSTFIX "d")`
`endif (MSVC)`

dcoeurjo: maybe.. but what's wrong for windows without ?

raphaellenain: On Windows, when I "cmake" DGtal for Visual Studio, the two configurations (Release and Debug) are automatically generated. Under Visual Studio, I can choose the configuration to build and install the library. After installing (make install equivalent), the file C:\Program Files\DGtal\lib\DGtal.lib is created (either in Debug or Release configuration, it's the same thing) (1).

Next to that, when I create a new project on Visual Studio, both of the configurations are automatically created for this new project, and both are useful (one for releasing the project, the other one for debugging) (2).

If my new project references DGtal, on the linking step, the Microsoft compiler works with C:\Program Files\DGtal\lib\DGtal.lib.
But if I created DGtal.lib with the Release configuration of DGtal (in step (1)), my new project, in its Debug configuration, won't be able to link with these library. 
On the other hand, if I created DGtal.lib with the Debug configuration of DGtal (in step (1)), my new project, in its Release configuration, won't be able to link with these library.

The debug version of DGtal.lib (size : 20 Mo) contains more information than the release version (size : 3 Mo), and the configuration (Debug or Release) is tagged on the windows static library files.

I hope my explanation is clear ...

raphaellenain: That's why I suggest adding a DGtald.lib for the Debug configuration

dcoeurjo: This is weird... on linux/mac, you can build an external program in Debug, linking to a Release library and it is just fine. If you debug your code, you will not have access to the debug symbols in the lib (which was compiled in Release) but it is just fine, you can still compile / run / debug you own code.

You are saying that it is not the case in windows ?

raphaellenain: I'm doing some tests for an alternative solution (without changing DGtal Cmake file :-) ). But the 2 errors on Visual studio are "_ITERATOR_DEBUG_LEVEL" and "RuntimeLibrary" : "https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk2038?view=vs-2017"

dcoeurjo: You can edit the cmake if you want (only for MSVC). I don't undrestand why VC complains but not a big deal if that fixes the issue

raphaellenain: I succeeded compiling my project in Debug mode, referencing the Release version of DGtal. 
When creating a new project on Visual Studio, the IDE automatically : 
 - for the Debug configuration : adds a preprocessor macro "_DEBUG" (--> _ITERATOR_DEBUG_LEVEL = 2), and sets the property "UseDebugLibraries" to true (--> RuntimeLibrary : Debug)
 - for the Release configuration : adds a preprocessor macro "NDEBUG" (--> _ITERATOR_DEBUG_LEVEL = 0), and sets the property "UseDebugLibraries" to false (--> RuntimeLibrary : Release)

So compiling in Debug mode, linking to the Release version of DGtal, is possible, changing these configurations, but it's not the default configuration for Visual Studio.

raphaellenain: Ok, I add the "MSVC" test

dcoeurjo: 👌

dcoeurjo:
👌

raphaellenain:
Thank you for the code review

dcoeurjo:
thanks to you for the PR.
Merging

PR #1416, opened 2019-06-04
Making HyperRectDomain_(sub)Iterator random access iterators.

rolanddenis:
# PR Description

## Random access iterator
`HyperRectDomain_Iterator` and `HyperRectDomain_subIterator` now behave like random-access iterator but they are not strictly random-access iterators due to the fact they refer to an internal member (`myPoint`) instead to a data of an iterated container.

One goal was to be able to easily split a `HyperRectDomain` into equal ranges, e.g. for parallelization purpose:
```C++
template <typename TIterator>
SimpleConstRange<TIterator>
split_range(TIterator it_begin, TIterator it_end, std::size_t idx, std::size_t count)
{
    auto range_size  = std::distance(it_begin, it_end);
    auto begin_shift = (range_size*idx) / count;
    auto end_shift   = (range_size*(idx+1)) / count;
    return { it_begin + begin_shift, it_begin + end_shift };
}

template <typename TIterable>
SimpleConstRange<TIterator>
split_range(TIterable & iterable, std::size_t idx, std::size_t count)
{
    return split_range(iterable.begin(), iterable.end(), idx, count);
}

[...]

#pragma omp parallel
{
  for (auto const& pt : split_range(image.domain(), omp_get_thread_num(), omp_get_max_threads()))
    {
      // do something with the point, in parallel
    }
}
```
## Increasing iterators comparison performance
Comparing two iterators now result in a comparison of their positions in the iterated sequence instead of comparing the referenced points.

However, it implies that comparing two iterators from different sequences (different domain and/or dimensions) is now undefined but it is anyway an invalid use of iterators (and it triggers an error in debug mode).

Some performance comparisons with `GNU/9`:

| bench | master | this PR |
| --- | --- | --- |
| domain traversal | 0.160542 Gpts/s | 0.993497 Gpts/s |
| domain reverse traversal | 0.0996051 Gpts/s | 0.995102 Gpts/s |
| domain traversal using subRange | 0.293052 Gpts/s | 0.317036 Gpts/s |
| domain reverse traversal using subRange | 0.0149367 Gpts/s | 0.309228 Gpts/s |

and with `Clang/9`:

| bench | master | this PR |
| --- | --- | --- |
| domain traversal | 0.160545 Gpts/s | 0.700166 Gpts/s |
| domain reverse traversal | 0.111249 Gpts/s | 0.874403 Gpts/s |
| domain traversal using subRange | 0.290743 Gpts/s | 0.347756 Gpts/s |
| domain reverse traversal using subRange | 0.0136115 Gpts/s | 0.19353 Gpts/s |

## Updating test file
Updating test file to use Catch test framework. Also adding random-access feature tests and adding a domain traversal benchmark (using the hidden `[.bench]` tag).

# Checklist

- [ ] All continuous integration tests pass (Travis & appveyor) (when https://github.com/DGtal-team/DGtalTools/pull/355 will be merged)
- [x] Parallelized traversal example using OpenMP


rolanddenis:
This PR is ready to be reviewed.

The compilation failure in DGtalTools is due to an attempt to pass a dereferenced iterator on `HyperRectDomain` to a function accepting a mutable reference to a `Point` (without modifying it). It was previously working because the current `HyperRectDomain_Iterator` has a corresponding deference operator (see https://github.com/DGtal-team/DGtal/blob/master/src/DGtal/kernel/domains/HyperRectDomain_Iterator.h#L193) but it was a design flaw since it could allow to modify the referenced point which doesn't make sense for a `HyperRectDomain`.

Will be fixed in https://github.com/DGtal-team/DGtalTools/pull/355

dcoeurjo:
Impressive perfs! Thanks

dcoeurjo:
Wanna do the benchmark test using catch2 ;) ?

rolanddenis:
Yep, I will test that now :wink:

rolanddenis:
Actually, I will wait for PR #1419 before testing catch benchmarks :wink:

dcoeurjo:
#1419  merged.
Let's play a bit with the catch2 benchmark vs. google micro-bench which is in DGtal (optional deps).

dcoeurjo:
Long time ago, I split benchmarks and unit-tests in the cmake files. The main reason was to speed up the `make test`  skipping the benchmarks. We would probably need to reconsider all that stuff (and maybe add a doc on benchmarking with catch).

src/DGtal/kernel/domains/HyperRectDomain_Iterator.h
```diff
@@ -43,325 +43,294 @@
...
-      myreverse_iterator(iterator_type __x) : current(__x),
-      prev(current)
-  { --prev; }
+namespace DGtal
+{
+  /** @brief Reverse iterator for HyperRectDomain
+   *
+   * @tparam TIterator  Iterator type on HyperRectDomain
+   *
+   * @note we need this specific implementation of a reverse iterator instead
```
dcoeurjo:
```suggestion
   * @note We need this specific implementation of a reverse iterator instead
```

tests/kernel/testHyperRectDomain.cpp
```diff
@@ -308,88 +368,165 @@ bool testSTLCompat()
...
 
-  ++nb; nbok += range.rbegin() == range.rend() ? 1 : 0;
-  trace.info() << "(" << nbok << "/" << nb << ") Checking that rbegin() == rend() for sub-rang {1}" << std::endl;
+  Point a, b;
+  Domain domain;
+};
 
-  trace.endBlock();
-  return nb == nbok;
+TEST_CASE_METHOD( BenchDomain, "Benchmarking domain traversal", "[.bench]" )
```
dcoeurjo: I don't understand your benchmark framework. Haven't used the Catch2 micro-benchmarking for that?

I don't see these test cases in the `./testHyperRectDomain -l`. Is that normal ?

rolanddenis: The dot at the beginning of the tag makes it hidden, ie it is not executed during the normal test run. However, I think the tag was listed with the `-t` option but it is not the case (maybe since the new version of Catch?). Anyway, you can run it with the `[bench]` option...

dcoeurjo: I’ll try, thx

dcoeurjo: Indeed `[bench]` is hidden when doing a `-t`. what do you mean by "Anyway, you can run it with the [bench] option..." ?

dcoeurjo: (and what is the diff between `TEST_CASE` and `TEST_CASE_METHOD`? )

rolanddenis: > Indeed `[bench]` is hidden when doing a `-t`. what do you mean by "Anyway, you can run it with the [bench] option..." ?

The "anyway" was maybe not appropriate :sweat_smile: I mean, you can call this hidden tag this way:
```bash
./test/kernel/testHyperRectDomain [bench]
```
```

Comment (rolanddenis) [reply tests/kernel/testHyperRectDomain.cpp]: > (and what is the diff between `TEST_CASE` and `TEST_CASE_METHOD`? )
> 
> (no doc in the catch2 repo https://github.com/catchorg/Catch2/blob/master/docs/test-cases-and-sections.md)

It is explained here: https://github.com/catchorg/Catch2/blob/master/docs/test-fixtures.md
I used it to launch the test as a class method, thus avoiding to repeat the context creation (typedefs and domain definition).

Comment (dcoeurjo) [reply tests/kernel/testHyperRectDomain.cpp]:
```
./testHyperRectDomain [bench]
zsh: no matches found: [bench]
```

Comment (dcoeurjo) [reply tests/kernel/testHyperRectDomain.cpp]:
```
./testHyperRectDomain -t bench
Tags for matching test cases:
0 tags
```

rolanddenis: And with quotes around `[bench]`? It seems to be interpreted by your shell instead of being passed to the binary...

dcoeurjo: working with the double-quotes..

dcoeurjo: thx

rolanddenis: > I don't understand your benchmark framework. Haven't used the Catch2 micro-benchmarking for that?

And about the included micro-benchmark, I have made some tests before the summer and I had some inconsistent results that I was trying to understand. 

One other strange thing is the performance of `subRange` iterator compared to the "full" iterator...

rolanddenis:
I have added a Google benchmark version and put the benchmarks into separate files to ease the usage (without using hidden tags) and avoid conflict between Catch and Benchmark.

I have also fix some issues:
- remove default benchmark target in `tests/CMakeList.txt` that conflicts with the one created when `BUILD_BENCHMARKS` is activated
- building benchmarks that use Google framework only if `BUILD_BENCHMARKS` is activated.

BTW, beside the similar name of `BUILD_BENCHMARKS` and `WITH_BENCHMARK` options, it was not clear that `BUILD_TESTING` also needs to be activated so that to build the benchmarks. We should maybe (one day) move benchmarks into a separate folder...

rolanddenis:
I have also changed the way the `check` variable is modified during the domain traversal and it is now more consistent with the benchmarks that use Catch and Google frameworks (there was probably a unwanted optimization when only adding the last coordinate of the point). I will update the PR description...

dcoeurjo:
`We should maybe (one day) move benchmarks into a separate folder...`

Agree.. Good idea..

(thx for the edits)

rolanddenis:
And about Catch and Google benchmark frameworks, the first seems easier to use but the later has more functionalities, like extracting the timings, displaying the processed items/s, and adaptive number of iterations, ...

dcoeurjo:
Still a minor update of the documentation would be missing, can you please have a look ?

rolanddenis:
Yep, and I will add a // example, like for initializing/modifying an image in parallel!

dcoeurjo:
;)thx

rolanddenis:
Done

rolanddenis:
Thanks, we also need to merge https://github.com/DGtal-team/DGtalTools/pull/359 to fix the compilation with DGtalTools!

dcoeurjo:
Yup, I saw that

PR #1429, opened 2019-07-16
Add const directives to some curve estimator on shapes.

akrah:
# PR Description

Add const directives to:

- the shape parameter (pointer) of constructors
- the private member corresponding to the constructor parameter
- the methods using the private member without calling non-const method on it

Currently, a user defining a shape need to declare/pass to a method a non-const shape as parameter.

# Checklist


akrah:
All is ok, ready for review.

ChangeLog.md
```diff
@@ -20,6 +20,10 @@
     (David Coeurjolly [#1418](https://github.com/DGtal-team/DGtal/pull/1418))
     (Roland Denis [#1419](https://github.com/DGtal-team/DGtal/pull/1419))
 
+- *Geometry*
+  - Add const directives to some curve estimator on shapes.
```
dcoeurjo:
```suggestion
  - Add const directives to some curve estimators on shapes.
```

copyme:
In the `ParametricShapeArcLengthFunctor.h` I suggest to use C++11` delete`  for the forbidden part of the interface (I think anyways the default constructor is not forbidden as the comment suggest), and `default` for the destructor. Also, could not be `myShape` turn into a reference? It does not look like we really need to pass it as a pointer. 

I know, my comments have nothing to do with the commit so if you do not feel like they are interesting, just ignore them.

dcoeurjo:
I agree for the `delete`/`default`in DGtal. There are many places (pre-c++11) where we could and probably should fix them.  

Agre also for the point vs. ref

akrah:
I also agree and applied the changes.
I needed to propagate the changes to files using this class, especially the TrueGlobalEstimatorOnPoints that use a `ParametricShapeArcLengthFunctor` as member with the the default constructor.
What do you think about?

rolanddenis:
Does `ParametricShapeArcLengthFunctor` need to model a DGtal functor ?

If it is so, then you cannot use here a (const) reference because it forbids the use of the assignment operator (that is already defined btw).

akrah:
@rolanddenis I suppose that it should model a DGtal functor but only the class name imposes this.
Maybe we can remove the assignment operator that seems useless?
Moreover, if it is a functor, maybe we should add it in the functor namespace?

rolanddenis:
Ok, it doesn't seem to model any functor concept and is only used together with `TrueGlobalEstimatorOnPoints` so yes, you can use references and delete the assignment operator :wink:

It also seems to be appropriate to put it in the `functors` namespace (@dcoeurjo ?).

About `TrueGlobalEstimatorOnPoint`, deleting the default constructor is not mandatory (other than for clarity) since a user-defined constructor is provided.
BTW, we can set its copy constructor and assignment operator to `= delete` instead of putting it in the private section :wink:

akrah:
Ok for replacement by `= delete`.

I also seen the `ParametricShapeCurvatureFunctor` and `ParametricShapeTangentFunctor`, similar by their design to `ParametricShapeArcLengthFunctor` : I think that it could be interesstin to add the same modifications (reference member, `= delete` and  `functors` namespace): what do you think ?

dcoeurjo:
@akrah can you please have a look to DGtaltools issue ?

dcoeurjo:
ping  @akrah

dcoeurjo:
reping @akrah  ;)

akrah:
I finally continue this PR. This last commit contains the minimal changes about our discussion. I tried to simplify the use of local and global estimators but this imply to also modify `CompareLocalEstimator` and to broke the `CCurveLocalGeometricEstimator`, specifying that an Estimator must have a constructor without parameter.

Maybe this could be interesting to make a review of this DGtal part with another issue? What d you think?

akrah:
> @akrah can you please have a look to DGtaltools issue ?

@dcoeurjo Sorry for the delay, do you think to a specific issue?

akrah:
> @akrah can you please have a look to DGtaltools issue ?

@dcoeurjo Sorry for the delay, I look at this.

akrah:
The goal of the last commit consists in refactoring `TrueGlobalEstimatorOnPoints` to be similar to `TrueLocalEstimatorOnPoints`.

From this point, I think that it could be interesting to refactor these two classes (and probably the corresponding concepts) for the following reasons:

1. In `TrueLocalEstimatorOnPoints`, iterators `myBegin` and `myEnd` are totally useless because never used.
2. In `TrueGlobalEstimatorOnPoints`, iterators `myBegin` and `myEnd` are only used if you use `init(h,itb,ite,isClosed)` then `eval()`, that internally call `eval(myBegin,myEnd,isClosed)`. Or another possibility could be to have `init(h)` then call `eval(itb,ite,isClosed)`. For users of `TrueGlobalEstimatorOnPoints`, they would also have two code lines.

So I propose to change both estimators to only have `init(h)` and `eval(itb,ite)` methods.
But this imply to make old code using these estimators invalid with the next DGtal version.
@dcoeurjo @rolanddenis: do you think we could make this refactoring?

dcoeurjo:
All good for me. Please go ahead;)
thx

akrah:
Done.

The last commit mainly contains these two changes:

1. Delete the initialization of all estimators from the concepts. I think that the initialization could be different depending on the estimator. The concept only imposes the computation method.
2. The grid resolution parameter is now on the computation methods because only used in these methods. Note that this parameter has always a default value to 1.

src/DGtal/geometry/curves/estimation/CompareLocalEstimators.h
```diff
@@ -148,31 +152,32 @@ namespace DGtal
...
     compareVectors(FirstEstimator & aFirstEstimator,
        SecondEstimator & aSecondEstimator,
-       const ConstIterator &it)
+       const ConstIterator &it,
+       const double h = 1.)
     {
       ASSERT( aFirstEstimator.isValid());
       ASSERT( aSecondEstimator.isValid());
-      Quantity v1 = aFirstEstimator.eval(it), v2 = aSecondEstimator.eval(it);
+      Quantity v1 = aFirstEstimator.eval(it,h), v2 = aSecondEstimator.eval(it,h);
```
copyme: it,h -> it, h (space)

src/DGtal/geometry/curves/estimation/MLPLengthEstimator.ih
```diff
@@ -38,109 +38,68 @@
...
+  FaithfulPolygon fp( itb, ite ); 
+  std::vector<Point> points(fp.size());
+  fp.copyMLP( points.begin() );
+
+  typename std::vector<Point>::const_iterator i = points.begin();
+  typename std::vector<Point>::const_iterator j = i;
+  ++j;
+  const typename std::vector<Point>::const_iterator end = points.end();
+
+  for ( ; j != end; ++i, ++j )
```
copyme: not better to use the other syntax i.e., `for (auto it : points)`?

src/DGtal/geometry/curves/estimation/ParametricShapeArcLengthFunctor.h
```diff
@@ -122,20 +115,17 @@ namespace DGtal
...
       double n = v.norm(RealPoint::L_infty);
       unsigned int nbSamples = (unsigned int) ceil( n*100 );
 
       //computes the angles
-      double t = myShape->parameter( aFirstPoint );
-      double t2 = myShape->parameter( aSecondPoint );
-      return myShape->arclength (t,t2,nbSamples);
+      double t = myShape.parameter( aFirstPoint );
+      double t2 = myShape.parameter( aSecondPoint );
+      return myShape.arclength (t,t2,nbSamples);
```
copyme: missing spaces after commas.

src/DGtal/geometry/curves/estimation/ParametricShapeArcLengthFunctor.h
```diff
@@ -144,24 +134,21 @@ namespace DGtal
...
-      ASSERT(myShape);
-
       //determining nbSamples from the bounding box size of the shape
-      RealPoint v = myShape->getUpperBound() - myShape->getLowerBound();
+      RealPoint v = myShape.getUpperBound() - myShape.getLowerBound();
       double n = v.norm(RealPoint::L_infty);
       unsigned int nbSamples = (unsigned int) ceil( n*100 );
 
-      return myShape->arclength (0,2*M_PI,nbSamples);
+      return myShape.arclength (0,2*M_PI,nbSamples);
```
copyme: spaces

src/DGtal/geometry/curves/estimation/SegmentComputerEstimators.h
```diff
@@ -1239,12 +1238,12 @@ namespace DGtal
...
-	Vector v = ( *aDSS.begin() - *boost::prior(aDSS.end()) ); 
-	Value l = v.norm(Vector::L_2);
-	//result
-	return 1/( (l*l)/8 + 0.5 );  
+        typedef typename DSS::Vector Vector; 
+        //length
+        Vector v = ( *aDSS.begin() - *boost::prior(aDSS.end()) ); 
+        Value l = v.norm(Vector::L_2);
+        //result
+        return 1/( (l*l)/8 + 0.5 );
```
copyme: Is `l.norm()` always bigger than `0`? Also, maybe 1 -> 1. and 8 -> 8.

src/DGtal/geometry/curves/estimation/SegmentComputerEstimators.h
```diff
@@ -1267,15 +1266,15 @@ namespace DGtal
...
-	return 1.0/( (l*l)/(8*w) + w/2 ); 
+        typedef typename DSS::Vector Vector; 
+        //length
+        Vector v = ( *aDSS.begin() - *boost::prior(aDSS.end()) ); 
+        Value l = v.norm(Vector::L_2);
+        //width
+        Vector t( aDSS.b(), aDSS.a() );
+        Value w = 1.0 / v.norm(Vector::L_2); 
+        //result
+        return 1.0/( (l*l)/(8*w) + w/2 );
```
copyme: similar here.

src/DGtal/geometry/curves/estimation/TwoStepLocalLengthEstimator.ih
```diff
@@ -38,109 +38,83 @@
...
+
+    ++ii;
+  }
+  while ( (i != ite) && (ii != ite) );
+
+  if ( i != ite )
+    nbDirect++;
   
-  return (nbDiag*myWeightDiagonal + nbDirect*myWeightDirect)*myH;
+  return (nbDiag*myWeightDiagonal + nbDirect*myWeightDirect)*h;
```
copyme: spaces

tests/geometry/curves/estimation/testTrueLocalEstimator.cpp
```diff
@@ -101,22 +101,21 @@ bool testTrueLocalEstimator(const std::string &filename)
...
   trace.info() << "Current point = "<<*it<<std::endl;
   trace.info() << "Current point+15 = "<<*it2<<std::endl;
-  trace.info() << "Eval curvature (begin, h=1) = "<< curvatureEstimator.eval(it2)<<std::endl;
-  trace.info() << "Eval tangent (begin, h=1) = "<< tangentEstimator.eval(it2)<<std::endl;
-  trace.info() << "Eval length ( h=1) = "<< lengthEstimator.eval(it,it2)<<std::endl;
+  // TODO: All these estimations consider projections of points from grid curve c on ball.
+  //       These are not estimations on the grid curve c. Not sure that make sense.
+  trace.info() << "Eval curvature (begin, h=1) = "<< curvatureEstimator.eval(it2,1.)<<std::endl;
+  trace.info() << "Eval tangent (begin, h=1) = "<< tangentEstimator.eval(it2,1.)<<std::endl;
+  trace.info() << "Eval length (h=1) = "<< lengthEstimator.eval(it,it2,1.)<<std::endl;
```
copyme: spaces

src/DGtal/geometry/curves/estimation/ParametricShapeArcLengthFunctor.h
akrah: done

src/DGtal/geometry/curves/estimation/ParametricShapeArcLengthFunctor.h
akrah: done

src/DGtal/geometry/curves/estimation/SegmentComputerEstimators.h
akrah: I'm not sure, I just fixed indentation.
But I think we can only have 0 if we deal with a DSS a size 0, i.e. than the first element of `aDSS` is the same than the last.

src/DGtal/geometry/curves/estimation/SegmentComputerEstimators.h
akrah: Same answer.

src/DGtal/geometry/curves/estimation/TwoStepLocalLengthEstimator.ih
akrah: done

tests/geometry/curves/estimation/testTrueLocalEstimator.cpp
akrah: done

akrah:
By adapting the DGtalTools relative to this PR, I found some differences into DSS, FP and MLP changes. In particular, the [lenghtEstimators tool](https://dgtal-team.github.io/doctools-nightly/lengthEstimators.html) documentation proposes an example where there is the following differences between if using the current master od DGtal or this PR (green curve corresponds to result using the master, purple using this PR).

I investigated on the points used by this PR to compute lengths but I think that all is ok. I suppose this is relative to the introduction of deal with Circulators, but I didn't identify bug into the new code.Someone to help me to understand?

DSS_changes
FP_changes
MLP_changes

src/DGtal/geometry/curves/estimation/SegmentComputerEstimators.h
akrah: done

src/DGtal/geometry/curves/estimation/SegmentComputerEstimators.h
akrah: done

dcoeurjo:
Let me ping @troussil on this PR..

troussil:
I will have a look

troussil:
The code is indeed different between the master and this branch for these 3 classes. Unfortunately, output is also different. It's difficult to say which code is correct, but in any case, the new code is cleaner. Here are my suggestions : 
- For DSS, try to use method `.setMode("Truncate+1");` of the `GreedySegmentation` object in order to close the polygon, 
- For FP and MLP, try to use circulators (because your code now manage circulators) in the [lengthEstimators tool](https://dgtal-team.github.io/doctools-nightly/lengthEstimators.html)
- For all, you may also have a look at the polygons used for the length estimation, especially in case of circulators on closed curves. If they are correct, the length estimation should be correct too.

kerautret:
ping @akrah did you check the @troussil answer ? ;)

kerautret:
@akrah In order to advance in the merge (with the DGtaltools) what about to change circulator by the  orignal constIterator in the  lengthEstimators tool?  If you can test and it does not change the result we can merge this PR and address another issue to fix the  use of Circulator/Iterator  ?  (ping  also @troussil @dcoeurjo )

kerautret:
@akrah for travis is the iterator is replaced it should fix travis on DGtalTools no ?

dcoeurjo:
Dear all,
I've merged this branch to the rel1.1 PR to check the final build. 
If it's fine, I'll close this PR?

kerautret:
@dcoeurjo thanks I checked the tools pass on my computer with DGtal 1.1 So @akrah can just merge its branch and I will merge (I suppose tomorrow should be fine)

dcoeurjo:
👌

dcoeurjo:
Ok closing this one, every commits are in the rel1.1 branch.
thx @akrah, waiting for your edits on DGtalTOols

akrah:
Hi all,
I managed the PR of DGtalTools.
I didn't had time to investigate on the differences seen above but maybe can I create an issue with this considerations to continue the investigations when I (or anoher one) will have time?

dcoeurjo:
👌 , please open up a new issue with all the material (and link to this PR).

kerautret:
+1 for the issue.

akrah:
Issue added.

PR #1443, opened 2019-12-13
Testing Travis ci

dcoeurjo:
# PR Description

*your description here*

# Checklist


rolanddenis:
Just seen that the wrong language is used in Travis: Ruby instead of C++ ...

rolanddenis:
There is something with the `secure`: by removing it, it seems to works (see https://github.com/rolanddenis/DGtal/pull/11). I also removed the `sudo` but just because it is deprecated (see https://docs.travis-ci.com/user/reference/trusty/#container-based-infrastructure).

I was wondering if the `secure` option should be under another key (like `env`) or if it fails just because Travis fails to decrypt it (deleted key ?)?

dcoeurjo:
You're right... the `secure` was involved in the bug.. I think we don't need it anymore (was for doc deploy)

(closing this PR)

PR #1446, opened 2020-01-05
Removing C++11 checks

dcoeurjo:
# PR Description

cmake macros are used instead of the homemade checks.

# Checklist


cmake/Common.cmake
```diff
@@ -28,6 +28,13 @@ foreach(p LIB BIN INCLUDE DATA)
   endif()
 endforeach()
 
+
+
+# -----------------------------------------------------------------------------
+# CPP11 
+# -----------------------------------------------------------------------------
+set (CMAKE_CXX_STANDARD 11)
+
```
rolanddenis: Also add
```suggestion
set (CMAKE_CXX_STANDARD_REQUIRED TRUE)

```
to avoid fallback to previous standard version if C++11 is not available (see https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED)

cmake/DGtalConfigGenerator.cmake
```diff
@@ -11,15 +11,13 @@ export(TARGETS  DGtal FILE "${PROJECT_BINARY_DIR}/DGtalLibraryDepends.cmake")
 set(DGTAL_INCLUDE_DIRS "${PROJECT_SOURCE_DIR}/src" "${PROJECT_BINARY_DIR}/src")
 set(DGTAL_LIB_DIR "${PROJECT_BINARY_DIR}/src")
 set(DGTAL_CMAKE_DIR "${PROJECT_BINARY_DIR}")
+set(CMAKE_CXX_STANDARD 11)
```
rolanddenis: Also add
```suggestion
set(CMAKE_CXX_STANDARD_REQUIRED TRUE)

```
to avoid fallback to previous standard version if C++11 is not available (see https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED)

cmake/Common.cmake
dcoeurjo: thanks!

rolanddenis:
So, we also have to update DGtalTools? (cf the failing buid)

dcoeurjo:
Yep we should (ping @Kerautret)

kerautret:
> Yep we should (ping @kerautret)

Yes I look it!

kerautret:
I fact no need to change DGtalTools cmake, @dcoeurjo I think you forget to remove DGTalConfig.cmake.in , I pull request the change that should fix it (I hope):
https://github.com/dcoeurjo/DGtal/pull/28

dcoeurjo:
👌 thanks.

dcoeurjo:
it should work now.

dcoeurjo:
All good. Thank you guys

PR #1448, opened 2020-01-08
[WIP] Add SplitFunctions to split domains in hypercubes

phcerdan:
# PR Description

Given a domain defined by lower and upper bound points and
`requested_number_of_splits` implements two functions:
- computeSplits returns the number of splits that the
region can hold (lesser or equal than the requested)
- getSplit returns the ith split defined by the outputLowerBound and
outputUpperBound of the split region.

Implementation acquired from ITK: itkImageRegionSplitterMultidimensional.cxx
https://github.com/InsightSoftwareConsortium/ITK/blob/f971477cdacff22a861dfcbaccd89c0cf2755af7/Modules/Core/Common/src/itkImageRegionSplitterMultidimensional.cxx

# Checklist

### DIFF
--- a/src/DGtal/topology/CubicalComplex.ih
+++ b/src/DGtal/topology/CubicalComplex.ih
@@ -603,6 +603,7 @@ bool
 DGtal::CubicalComplex<TKSpace, TCellContainer>::
 belongs( Dimension d, const Cell& aCell ) const
 {
+  ASSERT( d <= dimension );
   CellMapConstIterator it = myCells[ d ].find( aCell );
   return it != myCells[ d ].end();
 }
--- a/src/DGtal/topology/CubicalComplex.ih
+++ b/src/DGtal/topology/CubicalComplex.ih
@@ -198,7 +198,7 @@ DGtal::CubicalComplex<TKSpace, TCellContainer>::
 fillData( Data data )
 {
   for ( Dimension d = 0; d <= dimension; ++d )
-    clearData( d, data );
+    fillData( d, data );
 }
 //-----------------------------------------------------------------------------
 template <typename TKSpace, typename TCellContainer>
--- a/src/DGtal/topology/CubicalComplex.ih
+++ b/src/DGtal/topology/CubicalComplex.ih
@@ -62,7 +62,7 @@ CubicalComplex()
 template <typename TKSpace, typename TCellContainer>
 inline
 DGtal::CubicalComplex<TKSpace, TCellContainer>::
-CubicalComplex( ConstAlias<KSpace> aK )
+CubicalComplex( Clone<KSpace> aK )
   : myKSpace( &aK ), myCells( dimension+1 )
 {
 }
@@ -603,9 +603,27 @@ bool
 DGtal::CubicalComplex<TKSpace, TCellContainer>::
 belongs( Dimension d, const Cell& aCell ) const
 {
-  ASSERT( d <= dimension );
   CellMapConstIterator it = myCells[ d ].find( aCell );
   return it != myCells[ d ].end();
+}
+//-----------------------------------------------------------------------------
+template <typename TKSpace, typename TCellContainer>
+inline
+bool
+DGtal::CubicalComplex<TKSpace, TCellContainer>::
+belongs( const PreCell& aCell ) const
+{
+  return belongs( TKSpace::PreCellularGridSpace::uDim( aCell ), aCell );
+}
+//-----------------------------------------------------------------------------
+template <typename TKSpace, typename TCellContainer>
+inline
+bool
+DGtal::CubicalComplex<TKSpace, TCellContainer>::
+belongs( Dimension d, const PreCell& aCell ) const
+{
+  if (!myKSpace->uIsValid(aCell)) return false;
+  return belongs(d, myKSpace->uCell(aCell) );
 }
 //-----------------------------------------------------------------------------
 template <typename TKSpace, typename TCellContainer>
--- a/tests/topology/testVoxelComplex.cpp
+++ b/tests/topology/testVoxelComplex.cpp
@@ -46,7 +46,7 @@
 #include "DGtal/kernel/BasicPointPredicates.h"
 #include "DGtal/topology/NeighborhoodConfigurations.h"
 #include "DGtal/topology/tables/NeighborhoodTables.h"
-// #include <DGtal/io/viewers/Viewer3D.h>
+#include <DGtal/io/viewers/Viewer3D.h>
 ///////////////////////////////////////////////////////////////////////////////
 
 using namespace std;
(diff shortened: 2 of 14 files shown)


src/DGtal/topology/CubicalComplex.ih
```diff
@@ -62,7 +62,7 @@ CubicalComplex()
 template <typename TKSpace, typename TCellContainer>
 inline
 DGtal::CubicalComplex<TKSpace, TCellContainer>::
-CubicalComplex( ConstAlias<KSpace> aK )
+CubicalComplex( Clone<KSpace> aK )
```
rolanddenis: Beware that `Clone` will allocate new `KSpace` (in many cases, e.g. when passing by const-reference) when `operator&` is applied below. Thus, it is now needed to free this memory in the destructor of `CubicalComplex`!

BTW, why changing the passing type?

src/DGtal/topology/CubicalComplex.ih
```diff
@@ -603,12 +603,30 @@ bool
 DGtal::CubicalComplex<TKSpace, TCellContainer>::
 belongs( Dimension d, const Cell& aCell ) const
 {
-  ASSERT( d <= dimension );
```
rolanddenis: Why removing this check ? BTW, why not `d < dimenson` instead :octocat:

src/DGtal/topology/CubicalComplex.ih
rolanddenis: In addition, if we actually need to copy the Khalimsky space, then we should store it by value instead by pointer.

phcerdan: I would love to have a copy.
I suffered the same problem than other classes had with `ConstAlias` (sorry I cannot find what class/commit changed from ConstAlias because the same problem) they make it impossible to construct the object in place (like emplacing_back to a vector).

src/DGtal/topology/CubicalComplex.ih
phcerdan: Debugging mistake, during development of the `belongs` function accepting `PreCell`. Reverting. Thanks!

> BTW, why not d < dimenson instead
I would say because `d` can be equal to `dimension`

phcerdan:
WIP: This branch should be splitted into three. The first, with all the functionality to split domains, the second adding more specific splitComplex functions, and the third one, implementing the thinnings in parallel.

The first two steps are completed, but the third, using the thinning in parallel, needs more work.
Note for myself: Some border blocks between domains are not skeletonized, possibly because they are set to fixed, when they shouldn't (Some block = the border blocks that are close to the domain border).

src/DGtal/topology/CubicalComplex.ih
rolanddenis: Yes but `d` starts at 0 or am I mistaken?

phcerdan: 0 is a pointel, 1 is a linel, 2 a surfel and 3 a voxel. So it can be 3.

rolanddenis: Ho yes, sorry, I forgot that! :octocat:

src/DGtal/topology/CubicalComplex.ih
rolanddenis: Do you have an example that fails ?

This works on my side:
```C++
using KSpace = DGtal::KhalimskySpaceND<3>;
using Point = KSpace::Point;
using CC = DGtal::CubicalComplex<KSpace>;

KSpace K();
K.init(Point(0, 0, 0), Point(512, 512, 512));

std::vector<C> v;
v.reserve(10); // Not actually needed
v.emplace_back(K);
```

but:
```C++
v.resize(10);
```
fails because of the lack of default constructor.

rolanddenis: BTW, we could also store Khalimsky space by copy in CubicalComplex but I am wondering if it was designed that way in the case of a more heavier implementation of a Cellular Grid, like the `CodedKhalimskySpaceND` from https://dgtal-team.github.io/doc-nightly/moduleCellularTopology.html#dgtal_ctopo_sec3 (@JacquesOlivierLachaud ?)

tests/topology/testVoxelComplex.cpp
```diff
@@ -1076,6 +1077,25 @@ TEST_CASE_METHOD(Fixture_X, "X DistanceMap", "[x][distance][thin]") {
         SECTION("asymetricThinning"){
             auto vc_new = asymetricThinningScheme<FixtureComplex>(
                     vc, selectDistMax, skelEnd<FixtureComplex>, verbose);
+            // SECTION( "visualize the thining" ){
+            //     int argc(1);
+            //     char** argv(nullptr);
+            //     QApplication app(argc, argv);
+            //     Viewer3D<> viewer(ks_fixture);
+            //     viewer.show();
```
copyme, line 1085: Should not commented code be removed before merging with the master?

tests/topology/testVoxelComplex.cpp
```diff
@@ -1110,10 +1130,95 @@ TEST_CASE_METHOD(Fixture_X, "X DistanceMap", "[x][distance][thin]") {
...
+            // auto & vc_new = out.merged_complex;
+            // SECTION( "visualize the thining" ){
+            //     int argc(1);
+            //     char** argv(nullptr);
+            //     QApplication app(argc, argv);
+            //     Viewer3D<> viewer(ks_fixture);
+            //     viewer.show();
+            //
+            //     viewer.setFillColor(Color(200, 200, 200, 100));
+            //     for ( auto it = vc_new.begin(3); it!= vc_new.end(3); ++it )
```
copyme, line 1207: same here

dcoeurjo:
I need to close this PR to apply the `bfg` cleanup.
@phcerdan could you please recreate the PR after the release ?
thx

src/DGtal/topology/CubicalComplex.ih
phcerdan: Just as a future reference. This is the commit (and comment) changing Alias to CountedPtr in Domains.

```diff
2013-10-31 00c92bf0b Update doc and concept accordingly. Update DigitalSetFromMap. (Jacques-Olivier Lachaud)

diff --git a/src/DGtal/kernel/sets/DigitalSetBySTLVector.h b/src/DGtal/kernel/sets/DigitalSetBySTLVector.h
--- a/src/DGtal/kernel/sets/DigitalSetBySTLVector.h
+++ b/src/DGtal/kernel/sets/DigitalSetBySTLVector.h
@@ -64,9 +74,4 @@
-   * Description of template class 'DigitalSetBySTLVector' <p> \brief
-   * Aim: Realizes the concept CDigitalSet by using the STL container
-   * std::vector.
-   *
-   * It thus describes a modifiable set of points within the given
-   * domain [Domain].
-   *
-   * @tparam TDomain a realization of the concept CDomain.
-   * @see CDigitalSet,CDomain
+    @since 0.7 Domains are now hold with counted pointers and no more
+    only aliased. The problem was related to returning sets with a
+    locally constructed domain. With CountedPtr, you are sure that the
+    domain remains valid during the lifetime of your set.

```

PR #1452, opened 2020-01-09
Add public function .data() to PointVector

### LINKED ISSUE #1399: Add public .data() method to PointVector
This will provide all the accesors already available of std::array: https://en.cppreference.com/w/cpp/container/array/data

https://github.com/DGtal-team/DGtal/blob/master/src/DGtal/kernel/PointVector.h#L138-L142
Const and non-const versions:

```cpp
const T* PointVector::data() const noexcept
{
return myArray.data();
}

T* PointVector::data() noexcept
{
return myArray.data();
}
```

What do you think?
It will make easier and faster to interact with other libraries accepting pointers.

I haven't checked if this will be compatible for other types of Containers (used in the library) different than the default std::array.

phcerdan:
# PR Description

Adds data() to PointVector that returns myArray.data()

This allows to facilitate interoperability of PointVector in other
libraries where raw arrays are used.

PointVector already has functions begin(), rbegin(), that are tied
to the use of std containers. data() was missing.

Note that in c++17 data() is constexpr. But there is no value to add
the complexity of handling both cases in the definition.

Closes #1399

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -18,6 +18,9 @@
   - DGtal can be compiled and used as a project (git) submodule (David
   Coeurjolly [#1444](https://github.com/DGtal-team/DGtal/pull/1444))
   
+- *Kernel package*
+  - Add .data() function to PointVector to expose internal array data.
+    (Pablo Hernandez-Cerdan,[#1452](https://github.com/DGtal-team/DGtal/pull/1452))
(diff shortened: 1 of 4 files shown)


phcerdan:
> Changelog entry ?blush

Added! :)

PR #1469, opened 2020-03-17
Adding suggested `skip_cleanup: true` in deployment conf

rolanddenis:
To avoid stashing just generated documentation files

It was previously removed because of a warning that is was deprecated,
we will see...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,3 +78,4 @@ jobs:
        verbose: true
        local_dir: "$TRAVIS_BUILD_DIR/build/html"
        strategy: api
+       skip_cleanup: true


rolanddenis:
Self merging...

rolanddenis:
@dcoeurjo Seems to work now! Travis warns me about the deprecated `skip_cleanup` as it should be defaulted to `true` in new dpl v2 but it is not the case...

rolanddenis:
BTW the deploy strategy doesn't change anything...

dcoeurjo:
🎉🎉

dcoeurjo:
Thx

PR #1484, opened 2020-04-26
Itk image origin fix

kerautret:
# PR Description

Fix origin image that was not taken into account in class ImageContainerByITKImage.
(all ITK images were always shifted with origin (0,0,0))

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -110,6 +110,11 @@
   - Removing a `using namespace std;` in the Viewer3D hearder file. (David
     Coeurjolly [#1413](https://github.com/DGtal-team/DGtal/pull/1413))
 
+- *images*
+  - Fix origin image that was not taken into account in class
+    ImageContainerByITKImage. (Bertrand Kerautret
+    [#1482](https://github.com/DGtal-team/DGtal/pull/1482))
+
 - *Shapes package*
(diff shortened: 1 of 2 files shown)


ChangeLog.md
```diff
@@ -106,10 +106,16 @@
     CCW ordering by default (in 3D).
     (Jacques-Olivier Lachaud,[#1421](https://github.com/DGtal-team/DGtal/pull/1445))
 
+- *images*
+  - Fix origin image that was not taken into account in class
```
dcoeurjo, line 110:
```suggestion
  - Fix the image origin that was not taken into account in class
```

kerautret: thanks for suggestions ;) !

phcerdan:
I am getting crashes because of this, and I fear this is partly wrong. `GetPixel` expects an index, not a physical point (origin).

I think the fixing in lower and upper bound with origin is correct, but not when using Get/Set Pixel.

Before this patch, they were in the Index Space, after this patch, there is a mixture between Index Space and Physical Space.
The ImageRegion iterator iterate over indices, so it doesn't make sense to add origin there.

However, it is right that something was missing, I think it would be solved if we change `GetOrigin` for `GetLargestPossibleRegion().GetIndex()`.

PR Incoming

PR #1488, opened 2020-05-10
Fix bug in VoxelComplex when cell is at boundary of kspace.

phcerdan:
Using KSpace::PreCell in VoxelComplex mask methods.
Add an override to `belongs` method to work with precells.
This belongs returns false if cell is not valid in current kspace
instead of an error.

Adds internal const correctness to various methods.

Also corrects minor style issues.

# Checklist


phcerdan:
This change is cherry-picked from #1448

phcerdan:
Rebased on top of master. Let me know if I can help to move this forward.

dcoeurjo:
thanks @phcerdan

dcoeurjo:
all good.. merging.

PR #1489, opened 2020-05-15
Avoid setting globally feenableexcept in Linux when in Debug mode

phcerdan:
Discovered while using DGtal in a third party project, along other
libraries.
When DGtal library is loaded imposes `feenableexcept`.

`Common.cpp`:

```cpp
void beforeMain (void) __attribute__((constructor));
void beforeMain (void)
{
  feenableexcept ( FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW );
}
```

There are other libraries that relies on feenableexcept not enabled.
For example ITK reading .nrrd files.

Add an option `DGTAL_ENABLE_FLOATING_POINT_EXCEPTIONS`, OFF by default,
to control this behavior at compile time.

# PR Description

*your description here*

# Checklist

- [] No warning raised in Debug ```cmake``` mode (otherwise, Travis C.I. will fail).

### DIFF
--- a/.travis/build_all.sh
+++ b/.travis/build_all.sh
@@ -9,6 +9,7 @@ BTYPE="$BTYPE -DBUILD_EXAMPLES=false -DBUILD_TESTING=true"
 BTYPE="$BTYPE -DCMAKE_BUILD_TYPE=Debug -DWITH_MAGICK=true -DWITH_GMP=true\
               -DWITH_FFTW3=true -DWARNING_AS_ERROR=ON -DCMAKE_BUILD_TYPE=Debug \
               -DWITH_HDF5=true -DWITH_CAIRO=true -DWITH_QGLVIEWER=true -DWITH_EIGEN=true\
+              -DDGTAL_ENABLE_FLOATING_POINT_EXCEPTIONS=true \
               -DWARNING_AS_ERROR=OFF -DCMAKE_INSTALL_PREFIX='$DEPS_DIR'"
 
 # OSX dependencies and build customization (hence the source)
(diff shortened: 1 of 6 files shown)


phcerdan:
Replaced `add_compile_definitions` (require CMake 3.12) for `add_definitions`

copyme:
@phcerdan In fact `feenableexcept` should be turned on only for the testing module. Also, what kind of issues did you notice with ITK and `feenableexcept` being enabled? That sounds wired — maybe a bug in ITK?

copyme:
@phcerdan Please, provide an example reproducing the behavior of DGtal and ITK with  `feenableexcept`. Before, simply ignoring the issues by adding a switch I want to know what exactly is going on. At this point, it looks more like a bug somewhere than an incompatibility of ITK with `feenableexcept`.

copyme:
@phcerdan I suspect you are getting exceptions that do not make much sense at first, but from experience, I know that they are always bugs.

phcerdan:
The problem arises in a run-time sanity check in the ITK NrrdIO module (that uses a third-party library). There, there are "sanity checks" expecting that when you multiply two big numbers numbers you get the representation of a `inf` number. The problem is that multiplying two big numbers when `feenableexcept(FE_OVERFLOW)` raises an exception.
The checks are indeed questionable, but I don't think DGtal should toggle `feenableexcept` globally in any case.

The error raises here:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/NrrdIO/src/NrrdIO/miscAir.c#L52
Called from:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/NrrdIO/src/NrrdIO/sane.c#L82

To reproduce it, read a `nrrd` image in linux, with `CMAKE_BUILD_TYPE=Debug` in an executable linking with DGtal and ITK libraries (or with DGtal with the option `WITH_ITK`).

copyme:
@phcerdan Ah I see now. Well, then your changes have of course sense. Maybe, `feenableexcept` should not be enabled at all when DGtal is compiled with third-party libraries like ITK.

Thanks for the explanation!

phcerdan:
Is there something I can do to move this forward?

dcoeurjo:
I'm on it, thx for the PR

dcoeurjo:
waiting for travis but all good

PR #1492, opened 2020-05-21
Itk shift domain in reader

kerautret:
# PR Description

In relation to previous @phcerdan PR #1490 ,  it adds the possibility to import ITK images with a shifted domain in ITKReader. By default the parameter shiftDomainUsingOrigin was set to true in order to be consistent with importing a .vol file that also has a shifted domain by default.
It was well tested by displaying from the DGtalTools (3dImageViewer) test images.

# Checklist


src/DGtal/io/readers/ITKReader.ih
```diff
@@ -188,8 +188,29 @@ namespace DGtal {
...
+    if (shiftDomainUsingOrigin)
+    {
+      const auto c = itk_image->GetOrigin();
+      typename Domain::Point p;
+      for (unsigned int k = 0; k < dimension; k++)
+      {
+        p[k] = static_cast<typename Domain::Integer>(c[k]);
+      }
+      image.updateDomain(p);
+      //updating only the domain is not sufficient to import correctly
```
phcerdan, line 201: I would like to reproduce it. Is this exercised in the test?
I mean, if I remove this code, does the test fail?

Maybe we can fix it in the `operator()` of the ImageContainerByITKImage.

kerautret: yes I just complete the test in last commit, you can try to comment these lines and try ./tests/io/readers/testITKReader it will fails with the range() access to image value.

kerautret: If you have the DGtalTools, you can also try with and without these lines and you will have:
Capture d’écran 2020-05-22 à 21 13 04
And if you comment the lines you will have a shifted display:
Capture d’écran 2020-05-22 à 21 12 40

kerautret: > Maybe we can fix it in the operator() of the ImageContainerByITKImage.
I try to do some edits on the iterator but it brakes your tests.

phcerdan:
Hi @kerautret, I have opened a PR to your fork: https://github.com/kerautret/DGtal/pull/69

It changes the operator() and setValue() in ImageContainerByITK. ImageContainerByITKImage test is accommodated to these new changes.
I think with that, everyone should be happy.

The patch removes the change in the image region of ITK that you had to do.

kerautret:
Thank @phcerdan I look it!

phcerdan:
I added a minor patch as well. Changed the getPhysicalPointFromIndex to getPhysicalPointFromDomainPoint. To facilitate getting physical points from DGtal when the domainShift is not zero.

kerautret:
Hi @phcerdan, Thanks,  I merged your PR and all is nice! 🎉 I just add you in the changelog ;)

kerautret:
Perfect and thanks too @phcerdan for the interactions ! On my tests all looks fine even with other tools and in combination (extraction and display): 
Capture d’écran 2020-05-23 à 23 06 32

PR #1496, opened 2020-05-26
Add WITH_DGtalLUT option

phcerdan:
`DGtalLUT` is a library that pre-loads all the LookUpTables
of NeighborhoodTables.

The current mechanism in NeighborhoodTables.h.in to declare strings
with hard-coded paths pointing to the tables (the tables are stored
 in `.zlib` files in the repository) depending if we are
 in BUILD or INSTALL tree of DGtal works. However doesn't really
 allow re-usability of DGtal in third-parties projects.
Because, those strings will still be hard-coded, but the tables might
not be there.

An option would be to provide a `CMake` script to re-allocate the tables,
and modify the header with the hard-coded strings to the new locations.

However, a cleaner mechanism is to provide another library: DGtalLUT, with
the tables (the tables are of type: `dynamic_bitset`) pre-loaded.
This way, third-parties can just link to DGtalLUT and not worry about
re-allocating the zlib files.

A DGtalLUT library compiled as static weights 2MB ( 9 tables ).
When testing and `WITH_DGtalLUT` are enabled, the new way of setting
the table is exercised in `testVoxelComplex.cpp`

# Checklist


phcerdan:
Nevermind, this doesn't solve the problem, I would need to distribute the tables and change the header.

PR #1521, opened 2020-10-07
Add SplitFunctions to split domains in hypercubes

phcerdan:
Given a domain defined by lower and upper bound points and
requested_number_of_splits implements two functions:
- computeSplits returns the number of splits that the
region can hold (lesser or equal than the requested)
- getSplit returns the ith split and the outputLowerBound and
outputUpperBound of the split region.

Implementation acquired from ITK: itkImageRegionSplitterMultidimensional.cxx
https://github.com/InsightSoftwareConsortium/ITK/blob/f971477cdacff22a861dfcbaccd89c0cf2755af7/Modules/Core/Common/src/itkImageRegionSplitterMultidimensional.cxx

- Add splitComplex to SplitFunctions
- Add test and change testSplitFunctions to use Catch

# Checklist


phcerdan:
This is a rebase from https://github.com/DGtal-team/DGtal/pull/1448. The thinning algorithm with splits was discarded.

PR #1524, opened 2020-10-15
Modernize CMake

phcerdan:
- Avoid global `include_directories`, `link_libraries` etc. Use
instead the `target_` commands.
- Create DGtal library in the top directory, allowing to be targeted in
the other CMake files.
- Use lower case for CMake commands, i.e. `SET` to `set` etc.
- Remove naming the condition in `else`, `endif`, and `endforeach`: `if(x), else(x), endif(x)`
- Add function `DGtal_add_test(test_file)` to avoid repetition when
configuring tests. It accepts a dummy optional parameter, for example
`ONLY_ADD_EXECUTABLE` that will create the executable, link to DGtal,
but avoid the `add_test` command. Used mainly when benchmarking is ON.

# Checklist


dcoeurjo:
I am now a great fan of "Fact_content" (or any higher level facade like https://github.com/TheLartians/CPM.cmake). 
What do think to have the following deps "fetched" at cmake by default: 
* Eigen 
* GMP/GMPXX
* CGAL 

(with flag to allow to use the installed deps)

@JacquesOlivierLachaud ? @rolanddenis

rolanddenis:
Since CGAL is now header only, it is a good idea to simplify DGtal installation! If we replace `boost/program_options` by CL11 in DGtal, I suppose Boost should also be header only?

JacquesOlivierLachaud:
I am not a great expert on configuration tools. Everything that makes the life of future developers/users easier is good for me.
If we have to fetch by default some packages, the three you list is a good option (Eigen very useful but pulls a lot of other packages, gmp/gmpxx is a small dependency, CGAL is sometimes useful and header only).

dcoeurjo:
In DGtal, boost is already header only (only libs have been considered in DGtalTools but not true anymore I guess). It could also be "fetched" at cmake step but my only concern would be the size of the working copy (minor issue, though...).

Maybe the "make install" step is also something  that could introduce bugs..

phcerdan:
FetchContent uses `add_subdirectory`, so you end owning the third-party library in your project, i.e gmp.so would be at the same level than DGtal.so. If the CMake files of the third-party are not pretty modern, you might face global changes on `CMAKE_CXX_FLAGS`, or if missing alias targets, you might find incompatibilities of library names between `find_package` and `add_subdirectory`.

An alternative that from my experience works great is to have a `dependencies` subproject inside DGtal (but completely independent), that uses FetchContent to feed all the dependencies (with options). It will be built in its own folder, for example `build-dgtal-dependencies`.

With this approach, DGtal won't be changed, it will still expect `Eigen_DIR`, `CGAL_DIR` and use `find_package`. Then, a wrapper can be created, such as:  `DEPENDENCIES_DIR=/path/build-dgtal-dependencies` to where the dependencies subproject was built, that provide all the dependencies `_DIR` with just one option.

The `dependencies` approach is like the super-build approach, but you don't have to config all the dependencies every time.

phcerdan:
> Since CGAL is now header only, it is a good idea to simplify DGtal installation! If we replace `boost/program_options` by CL11 in DGtal, I suppose Boost should also be header only?

I think boost-graph can enjoy being linked (reduces compile time).

phcerdan:
In any case, this modernization is about removing global commands for target specific commands. Remove upper case, and modernize `DGtalConfig.cmake` file accordingly.

dcoeurjo:
Additional things that should be in the Fetch_content stuff:
* Catch2 (for the tests)
* CLI11 (maybe DGtalTools only)
* LibBoard? (though the main project seems to be dead)

dcoeurjo:
I've just tried a first Fetch_content with Catch2...

cmake/FetchExternalDeps.cmake
```diff
@@ -1,5 +1,9 @@
 Include(FetchContent)
```
phcerdan, line 2: Maybe only fetch Catch2 if testing is enabled?
`if(BUILD_TESTING)`

dcoeurjo: oh yeah.. sure.. 👍

phcerdan:
@dcoeurjo  I will rebase on top of master instead of:  `Merge branch `master` into this branch`.
Are you working on something on this branch? The head of this branch will be invalidated

dcoeurjo:
please go ahead, I'll resest my working copy if I need.

dcoeurjo:
Can this branch be merged ? @phcerdan anything that is still WIP here?
We may consider other FecthContent in the future but maybe in a separate PR..

phcerdan:
> Can this branch be merged ?

I think yes.

> We may consider other FecthContent in the future but maybe in a separate PR..

Sounds good!

cmake/DGtalConfigGenerator.cmake
```diff
@@ -7,29 +7,39 @@ export(TARGETS  DGtal FILE "${PROJECT_BINARY_DIR}/DGtalLibraryDepends.cmake")
...
 
-# Create a DGtalConfig.cmake file for the use from the build tree
-set(DGTAL_INCLUDE_DIRS "${PROJECT_SOURCE_DIR}/src" "${PROJECT_BINARY_DIR}/src")
-set(DGTAL_LIB_DIR "${PROJECT_BINARY_DIR}/src")
-set(DGTAL_CMAKE_DIR "${PROJECT_BINARY_DIR}")
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 
+# Provide HINTS to find_dependency in DGtalConfig.cmake if DGTAL_CONFIG_HINTS
+# is ON. Disable it when deploying.
```
dcoeurjo, line 12: what is find_dependency ?

cmake/FindCOIN3D.cmake
```diff
@@ -7,14 +7,14 @@
 #
  
 
-IF (WIN32)
```
dcoeurjo, line 10: Is Coin3D still used in DGtal ? I don't think so.. will check later.

dcoeurjo (approved):
Excellent, I just have a minor question.

dcoeurjo:
Ok SoQT and COin3D can be removed.. I'll do it

dcoeurjo (changes requested):
+ Changelog entry?

cmake/DGtalConfigGenerator.cmake
phcerdan: Is like find_package, but with a better error message, if the package is missing. Only used in `FooConfig.cmake` files.

dcoeurjo: 👍

dcoeurjo (approved):

dcoeurjo (approved):
All good... thx

PR #1526, opened 2020-10-16
Only set CMAKE_CXX_STANDARD if not defined already

phcerdan:
This allow users to compile with their own standard.

# Checklist

### DIFF
--- a/cmake/Common.cmake
+++ b/cmake/Common.cmake
@@ -31,10 +31,21 @@ endforeach()
 
 
 # -----------------------------------------------------------------------------
-# CPP11 
+# CPP11
 # -----------------------------------------------------------------------------
-set (CMAKE_CXX_STANDARD 11)
-set (CMAKE_CXX_STANDARD_REQUIRED TRUE)
+set(DGTAL_CMAKE_CXX_STANDARD_MIN_REQUIRED 11)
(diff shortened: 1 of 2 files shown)


cmake/Common.cmake
```diff
@@ -31,10 +31,14 @@ endforeach()
 
 
 # -----------------------------------------------------------------------------
-# CPP11 
+# CPP11
 # -----------------------------------------------------------------------------
-set (CMAKE_CXX_STANDARD 11)
-set (CMAKE_CXX_STANDARD_REQUIRED TRUE)
+if(NOT CMAKE_CXX_STANDARD)
```
dcoeurjo, line 36: is CMAKE_CXX_STANDARD always >= 11 ?

dcoeurjo (commented):
Thx @phcerdan fir the fix, just a minor question..

cmake/Common.cmake
phcerdan: Well, the user could explicitly provide `-DCMAKE_CXX_STANDARD=98`, bu t I have never seen that.
I can throw a fatal error if defined and less than 11, will update.

dcoeurjo (approved):

dcoeurjo:
Perfect, thanks @phcerdan

PR #1532, opened 2020-11-19
Add container member function to DigitalSets and ImageContainers

phcerdan:
# PR Description

`container()` returns a reference (might be const) to the underlying
container holding the data.
It might point to an instance of Self type, Parent type, or to a protected/private
data member, depending on the type.

# Checklist

### DIFF
--- a/src/DGtal/images/ImageContainerByHashTree.h
+++ b/src/DGtal/images/ImageContainerByHashTree.h
@@ -276,6 +276,21 @@ namespace DGtal
      */
     Range range() ;
 
+    /**
+     * Give access to the underlying container.
+     * @return a (might be const) reference to the container.
+     */
+    const Self & container() const { return *this; };
+    /** @copydoc container() */
+    Self & container() { return *this; };
+
+    /**
+     * Give access to the underlying data.
+     * @return a (might be const) reference to the data.
+    */
+    const Node** & data() const noexcept { return myData; };
+    /** @copydoc data() */
+    Node** & data() noexcept { return myData; };
 
     /**
      * Returns the value corresponding to a key.
--- a/src/DGtal/images/ImageContainerBySTLVector.h
+++ b/src/DGtal/images/ImageContainerBySTLVector.h
@@ -129,6 +129,8 @@ namespace DGtal
   public:
 
     typedef ImageContainerBySTLVector<TDomain, TValue> Self;
+    typedef std::vector<TValue> Parent;
+    typedef Parent Container;
 
     /// domain
     BOOST_CONCEPT_ASSERT ( ( concepts::CDomain<TDomain> ) );
@@ -283,6 +285,14 @@ namespace DGtal
      * iterators to scan the values of image.
      */
     Range range();
+
+    /**
+     * Give access to the underlying container.
+     * @return a (might be const) reference to the container.
+    */
+    const Container & container() const { return static_cast<Parent>(*this); };
+    /** @copydoc container() */
+    Container & container() { return static_cast<Parent>(*this); };
 
 
     /////////////////////////// Custom Iterator ///////////////
--- a/src/DGtal/kernel/sets/DigitalSetByAssociativeContainer.ih
+++ b/src/DGtal/kernel/sets/DigitalSetByAssociativeContainer.ih
@@ -325,6 +325,22 @@ DGtal::DigitalSetByAssociativeContainer<Domain, Container>::end()
   return mySet.end();
 }
 
+template <typename Domain, typename Container>
+inline
+const typename DGtal::DigitalSetByAssociativeContainer<Domain, Container>::Container &
+DGtal::DigitalSetByAssociativeContainer<Domain, Container>::container() const
+{
+  return mySet;
+}
+
+template <typename Domain, typename Container>
+inline
+typename DGtal::DigitalSetByAssociativeContainer<Domain, Container>::Container &
+DGtal::DigitalSetByAssociativeContainer<Domain, Container>::container()
+{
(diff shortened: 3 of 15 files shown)


dcoeurjo:
👍  
Shouldn't we update the Concepts as well if we rely on this ?
At this point, it would only be required for the python binding, not the actual "algorithms" using these data structures. @JacquesOlivierLachaud ?

phcerdan:
Yeah, it's just to expose the container, I don't think we need to update the Concept, but I am all ears.
I could also make the pybind class a friend, but, I kind of prefer to not mix python in the pure C++ side.

src/DGtal/images/ImageContainerByHashTree.h
```diff
@@ -276,6 +276,19 @@ namespace DGtal
      */
     Range range() ;
 
+    /**
+     * Give access to the underlying container.
+     * @return a (might be const) reference to the container.
+    */
+    inline const Self & container() const { return *this; };
+    inline Self & container() { return *this; };
```
JacquesOlivierLachaud, line 284: Stupid question. Doxygen allows to comment one method and extend the comment to the next method if  they have the same name (but different signature) ?

src/DGtal/images/ImageContainerByHashTree.h
```diff
@@ -276,6 +276,19 @@ namespace DGtal
...
+
+    /**
+     * Give access to the underlying data.
+     * @return a (might be const) reference to the data.
+    */
+    inline const Node** & data() const noexcept { return myData; };
+    inline Node** & data() noexcept { return myData; };
```
JacquesOlivierLachaud, line 291: Same question here and at everyplace afterwards.

src/DGtal/images/ImageContainerByHashTree.h
```diff
@@ -276,6 +276,19 @@ namespace DGtal
      */
     Range range() ;
 
+    /**
+     * Give access to the underlying container.
+     * @return a (might be const) reference to the container.
+    */
+    inline const Self & container() const { return *this; };
```
JacquesOlivierLachaud, line 283: I think inline keyword is redundant with the code placed at the definition (ie inlined).
https://isocpp.org/wiki/faq/inline-functions

src/DGtal/kernel/sets/DigitalSetByAssociativeContainer.ih
```diff
@@ -325,6 +325,22 @@ DGtal::DigitalSetByAssociativeContainer<Domain, Container>::end()
   return mySet.end();
 }
 
+template <typename Domain, typename Container>
+inline
```
JacquesOlivierLachaud, line 329: inline keyword is pertinent here.

src/DGtal/images/ImageContainerBySTLVector.h
```diff
@@ -129,6 +129,7 @@ namespace DGtal
   public:
 
     typedef ImageContainerBySTLVector<TDomain, TValue> Self;
+    typedef std::vector<TValue> Parent;
```
JacquesOlivierLachaud, line 132: Should we set an identical typedef ... Container ?

src/DGtal/kernel/sets/DigitalSetFromMap.h
```diff
@@ -269,6 +269,13 @@ namespace DGtal
      */
     Iterator end();
 
+    /**
+     * Give access to the underlying container.
+     * @return a (might be const) reference to the stored container.
+    */
+    inline const Image & container() const { return *myImgPtr;};
```
JacquesOlivierLachaud, line 276: Should we add a typedef Image Container; ?

JacquesOlivierLachaud (approved):

JacquesOlivierLachaud:
The problem in Travis seems unrelated to your PR.

src/DGtal/images/ImageContainerByHashTree.h
phcerdan: I thought yes, but I was wrong. This is the doxygen generated in from the same scenario in [PointVector::data()](https://github.com/DGtal-team/DGtal/blob/master/src/DGtal/kernel/PointVector.h#L974-L980)

image

There is a `1/2`, `2/2` index in the upper corner, but the second method is not documented.
I will fix it.

phcerdan: However there is  the command `@copydoc data()` that should work.

src/DGtal/kernel/sets/DigitalSetFromMap.h
phcerdan: Added

src/DGtal/images/ImageContainerBySTLVector.h
phcerdan: Added

src/DGtal/images/ImageContainerByHashTree.h
phcerdan: Good to know, fixed

src/DGtal/images/ImageContainerByHashTree.h
phcerdan: See other comment in the same topic

src/DGtal/images/ImageContainerByHashTree.h
phcerdan: Added @copydoc to the second function.

phcerdan:
> The problem in Travis seems unrelated to your PR.

The travis errors in clang seems related to doxygen, opened #1534 about it

Issue #1542, opened 2021-01-11
Failing doxygen after CMake modernization


phcerdan:
```cmake
-- Found Eigen3: /home/travis/build/DGtal-team/DGtal/deps/include/eigen3 (Required is at least version "2.91.0") 
-- Eigen3 (version 3.3.7) found.
-- Found FFTW3: /usr/include  
-- Found FFTW3_FLOAT: /usr/include  
-- Found FFTW3_DOUBLE: /usr/include  
-- Found FFTW3_LONG: /usr/include  
-- FFTW3 is found : /usr/lib/x86_64-linux-gnu/libfftw3f.so;/usr/lib/x86_64-linux-gnu/libfftw3.so;/usr/lib/x86_64-linux-gnu/libfftw3l.so;/usr/lib/x86_64-linux-gnu/libfftw3f_threads.so;/usr/lib/x86_64-linux-gnu/libfftw3_threads.so;/usr/lib/x86_64-linux-gnu/libfftw3l_threads.so.
-- -------------------------------------------------------------------------------
-- Build test files DISABLED (you can activate unit tests with '-DBUILD_TESTING=ON' cmake option)
-- -------------------------------------------------------------------------------
-- Build examples ENABLED
-- -------------------------------------------------------------------------------
-- Generating DGtalConfig files
-- Configuring done
CMake Error in CMakeLists.txt:

  Target "DGtal" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/home/travis/build/DGtal-team/DGtal/deps/include/eigen3"

  which is prefixed in the source directory.
-- Generating done
-- Build files have been written to: /home/travis/build/DGtal-team/DGtal/build
```

PR #1543, opened 2021-01-11
Modernize cmake: Prefer using targets. Fix CMake warnings

phcerdan:
# PR Description

Prefer using targets for Eigen and ZLIB:
- Require at least Eigen3.3 to use Eigen3Config.cmake and Eigen3::Eigen
target. Remove unused file: FindEigen3.cmake after this change. 
- Use target ZLIB::ZLIB instead of target_include_directories and
libraries.
- Also, because Boost and Zlib are required dependencies, remove the
`if(Boost_FOUND)` conditional.

CMake: Fix policy warning CMP0048 about project not handling VERSION:

- CMake does not allow yet not numeric versions, (DGtal use `beta` in version)
so we just suppress the policy warning setting it to OLD behavior.

CMake: Remove linking with DGtalLibDependencies in examples:

- DGtalLibDependencies is a list of libraries, but not needed since
recent modernization of CMake files. The target `DGtal` contains those libraries by itself.

# Checklist


phcerdan:
This solves the failing travis-ci with doxygen Fix #1542

ChangeLog.md
```diff
@@ -7,6 +7,8 @@
     (Pablo Hernandez-Cerdan [#1529](https://github.com/DGtal-team/DGtal/pull/1529))
   - Modernize CMake: Avoid global includes and links, use `target_` commands instead
     (Pablo Hernandez-Cerdan, David Coeurjolly [#1524](https://github.com/DGtal-team/DGtal/pull/1524))
+  - Modernize CMake: Prefer use targets than directories and libraries
```
dcoeurjo, line 10:
```suggestion
  - Modernize CMake: Prefer use targets rather than directories and libraries
```

dcoeurjo (changes requested):
Fine for me, thanks @phcerdan ! Just a minor comment for the changelog

ChangeLog.md
phcerdan: Thanks, fixed. I also removed linking to the list of libraries `${DGtalLibDependencies}` in examples. Those libraries are in the DGtal target already.

dcoeurjo:
Perfect. Merging.
I would need to fix the pending azure hook..

dcoeurjo (approved):

PR #1580, opened 2021-06-14
Dockerfile

msalazarcgeo:
# PR Description

*Dockerfile added*

# Checklist

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,4 +1,8 @@
 # DGtal 1.3(beta)
+- *Docker File*
+   - A Dockerfile is added to create a Docker image with the purpose of having a base
+     to start a development using the DGtal library.
+	
 
 
 # DGtal 1.2
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,7 +1,7 @@
 # DGtal 1.3(beta)
 - *Docker File*
-   - A Dockerfile is added to create a Docker image with the purpose of having a base
-     to start a development using the DGtal library.
+    - A Dockerfile is added to create a Docker image to have a base to start development
+      using the DGtal library.
 	
 
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,7 +1,9 @@
 # DGtal 1.3(beta)
+
+## New features / critical changes
 - *Docker File*
-    - A Dockerfile is added to create a Docker image to have a base to start development
-      using the DGtal library.
+  - A Dockerfile is added to create a Docker image to have a base to start development
+    using the DGtal library.
 	
 
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -3,7 +3,7 @@
 ## New features / critical changes
 - *Docker File*
   - A Dockerfile is added to create a Docker image to have a base to start development
-    using the DGtal library.
+    using the DGtal library.(J. Miguel Salazar [#1580](https://github.com/DGtal-team/DGtal/pull/1580)) 
 	
 
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ DGtal can be compiled on Microsoft Windows system using Visual Studio 2014 (or n
 Docker
 ======
 
-A Docker file is included in the repository in order to allow the users to build a docker image with the Dgtal library installed. Inside the docker container, the Dgtal library is installed so the user can start developing without the need to install the DGtal library on the system.
+A Docker file is included in the repository in order to allow the users to build a docker image with the DGtal library installed. Inside the docker container, the library is installed with many dependencies so the user can start developing without the need to install DGtal on the system.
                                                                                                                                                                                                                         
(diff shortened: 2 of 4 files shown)


ChangeLog.md
```diff
@@ -1,7 +1,7 @@
 # DGtal 1.3(beta)
 - *Docker File*
-   - A Dockerfile is added to create a Docker image with the purpose of having a base
-     to start a development using the DGtal library.
+    - A Dockerfile is added to create a Docker image to have a base to start development
+      using the DGtal library.
```
dcoeurjo: Could you please add your name + link to this PR

README.md
```diff
@@ -49,6 +49,23 @@ Minimum system requirements: C++11 enabled compiler, [cmake](http://cmake.org),
 
 DGtal can be compiled on Microsoft Windows system using Visual Studio 2014 (or newer): Generate the Visual Studio project using windows [cmake](http://cmake.org) tool and compile the DGtal solution (you may also need to set the cmake variable ```BUILD_SHARED_LIBS``` to false).
 
+
+Docker
+======
+
+A Docker file is included in the repository in order to allow the users to build a docker image with the Dgtal library installed. Inside the docker container, the Dgtal library is installed so the user can start developing without the need to install the DGtal library on the system.
```
dcoeurjo, line 56:
```suggestion
A Docker file is included in the repository in order to allow the users to build a docker image with the DGtal library installed. Inside the docker container, the library is installed with many dependencies so the user can start developing without the need to install DGtal on the system.
```

dcoeurjo:
thanks @msalazarcgeo for the PR.
Could you also please add a new line to the AUTHORS ?

msalazarcgeo:
> thanks @msalazarcgeo for the PR.
> Could you also please add a new line in the AUTHORS file?

Do yo want me to add my name to it?

ChangeLog.md
msalazarcgeo: Done in 6adb2b0c27fc059b2c9e1ac724ded8592109d7bb

dcoeurjo:
For the Qt thing (it would be great to have a solution for that), https://forums.docker.com/t/start-a-gui-application-as-root-in-a-ubuntu-container/17069 or http://fabiorehm.com/blog/2014/09/11/running-gui-apps-with-docker/ seem to be related...

dcoeurjo:
(thx for the AUTHORS)

msalazarcgeo:
> For the Qt thing (it would be great to have a solution for that), https://forums.docker.com/t/start-a-gui-application-as-root-in-a-ubuntu-container/17069 or http://fabiorehm.com/blog/2014/09/11/running-gui-apps-with-docker/ seem to be related...


As  I mentioned when I send you the file, I was able to run the GIU on my computer using what you have found (sharing the X-server files). But I couldn't run it without the Nvidia driver since my system has only an Nvidia-GPU as an output video. To connect your system with the docker container using the "digital" user you have to add the following share volumes to the docker run command 

```
-v /tmp/.X11-unix:/tmp/.X11-unix:rw
 --privileged -h $HOSTNAME -v $HOME/.Xauthority:/home/digital/.Xauthority
```

This would allow the "digital" user to run most of the GUI (that is the case on my system with the Nvidia drivers). I couldn't test on another computer my apologies.

dcoeurjo:
Thanks for the comment (I'm not specialist in Docker). I'll try on a Nvidia machine (even if I'm not sure to so why it is nvidia GPU specific but anyway..).

msalazarcgeo:
> Thanks for the comment (I'm not specialist in Docker). I'll try on a Nvidia machine (even if I'm not sure to so why it is nvidia GPU specific but anyway..).

If you try this image on a machine with an Nvidia GPU I guess that it will not work. If you want, I can send you the Docker file that I use. The reason why I believe that It won't work is that the Docker image does not have the Nvidia drivers installed.

These parameters 
```
-v /tmp/.X11-unix:/tmp/.X11-unix:rw
 --privileged -h $HOSTNAME -v $HOME/.Xauthority:/home/digital/.Xauthority
```
don't have anything to do with the fact that have if you have an Nvidia GPU. The parameters are added so that the Docker container has access to the X11 server and the Xauthority file on the system (with the proper credentials) and that the docker container can send the generated image to the X11 server to be displayed (or at least that is what I understand, I'm not an expert neither).

dcoeurjo:
Sorry @msalazarcgeo for the delay. I was in a rush this week and would have more time the next one.

Just a quick question: are all optional deps enabled in your docker image ?

msalazarcgeo:
> Sorry @msalazarcgeo for the delay. I was in a rush this week and would have more time the next one.
> 
> Just a quick question: are all optional deps enabled in your docker image ?

Don't worry @dcoeurjo.

*"Just a quick question: are all optional deps enabled in your docker image ?"*

 Not all, but I tried to put as many as I could. In the Dockerfile you can see the cmake command to know the ones that are up.

dcoeurjo:
I think we're good.  thx @msalazarcgeo !
🎉

PR #1607, opened 2021-10-01
bug fix in initialization of plane-probing computers on to digital surfaces

troussil:
# PR Description

There was a bug in ExamplePlaneProbingSurfaceLocalEstimator (when built in Debug mode). It was caused by a bad validity check and a corner case in the initialization step, because a computer may not be always initialized (See method 'eval' in PlaneProbingDigitalSurfaceLocalEstimator). ExamplePlaneProbingSurfaceLocalEstimator can now be run without error. 

# Checklist

(Not relevant for a bug fix)

- [ ] No warning raised in Debug ```cmake``` mode (otherwise, C.I. will fail).
- [ ] All continuous integration tests pass (Github Actions & appveyor)

### DIFF
--- a/src/DGtal/geometry/surfaces/estimation/PlaneProbingDigitalSurfaceLocalEstimator.h
+++ b/src/DGtal/geometry/surfaces/estimation/PlaneProbingDigitalSurfaceLocalEstimator.h
@@ -309,9 +309,10 @@ namespace DGtal
      *
      * @param aInitialFrame an initial probing frame.
      * @param aPreEstimation a pre-estimation vector.
-     * @return aInitialFrame if no frame were found, the new frame otherwise.
+     * @return a pair (false, aInitialFrame) if no frame was found, 
+     * a pair (true, newFrame) if a frame 'newFrame' was found. 
      */
-    ProbingFrame probingFrameWithPreEstimation (ProbingFrame const& aInitialFrame, RealPoint const& aPreEstimation) const;
+    std::pair<bool, ProbingFrame> probingFrameWithPreEstimation (ProbingFrame const& aInitialFrame, RealPoint const& aPreEstimation) const;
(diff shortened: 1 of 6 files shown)


ChangeLog.md
```diff
@@ -13,6 +13,10 @@
   - Continuous integration does not use Travis anymore but Github
     Actions. (David Coeurjolly, [#1591](https://github.com/DGtal-team/DGtal/pull/1591))
 
+- *Geometry package*
+  - Bug fix in ExamplePlaneProbingSurfaceLocalEstimato.
```
dcoeurjo, line 17:
```suggestion
  - Bug fix in examplePlaneProbingSurfaceLocalEstimator.
```

dcoeurjo (approved):
There is more than just a bug fix in the example. Maybe you should comment on the edits you have done in the main classes as well in the Changelog. 
Otherwise, everything is ok and the example does not fail anymore, thx

ChangeLog.md
troussil: Actually, there are not so much changes. The diff gives a somewhat erroneous impression. The most significant changes are: 

- in `PlaneProbingDigitalSurfaceLocalEstimator`
  - `probingFrameWithPreEstimation` now returns a pair bool-frame instead of just a frame, in order to tell whether the frame will lead to a valid initialization or not. 
  - `eval` now uses this boolean value to return the trivial normal vector if it has been set to 'False' (which means that no valid initialization is possible)
- in `PlaneProbingParallelepipedEstimator`: `isValid` does not call the `isValid` method of the delegate, but only checks the relevant parts (which have been pushed in to separate methods). 

In addition, there are a few cosmetic changes: `InPlane` -> `inPlane` and some function definitions have been put in the .ih files.

PR #1638, opened 2022-03-21
Fix testLongvol and testCompressedVolWriter

rolanddenis:
# PR Description

Fix `testLongvol` and `testCompressedVolWriter` that was failing on some configurations.

Actually, it was due to a left-shift overflow (eg a 56 bits left-shift of a `char`) in the `read_word` method of `LongvolReader` and it was possible to make tests failing on more configurations by making tested value larger (see first commit).

# Checklist

- [x] All continuous integration tests pass (Github Actions & appveyor)

### DIFF
--- a/.github/workflows/buildAndDocumentation-PR.yml
+++ b/.github/workflows/buildAndDocumentation-PR.yml
@@ -9,7 +9,7 @@ on:
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE:   Release
-  TESTBLACKLIST: "(testCompressedVolWriter|testLongvol|testLinearStructure)"
+  TESTBLACKLIST: "(testLinearStructure)"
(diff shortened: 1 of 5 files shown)


dcoeurjo:
Nice, what about testCompressedVol ? Have you been able to reproduce the bug?

dcoeurjo (approved):
All good to me.. thanks !

rolanddenis:
> Nice, what about testCompressedVol ? Have you been able to reproduce the bug?

You mean `testCompressedVolWriter`? I don't find any `testCompressedVol` :thinking: 

BTW, the last CI fails because of an assertion on the computational time in `testPolygonalCalculus` that should be lower with cache enabled compared to without cache.

dcoeurjo:
> You mean `testCompressedVolWriter`? I don't find any `testCompressedVol` 🤔

my bad.. I removed my comment ;)

> BTW, the last CI fails because of an assertion on the computational time in `testPolygonalCalculus` that should be lower with cache enabled compared to without cache.

Let me check. thx anyway

Issue #1629, opened 2022-01-20
Bug in DGtal::functions::Hull2D::computeHullThickness 


aramadancs:
The function returns 0 for a convex hull calculated using MelkmanConvexHull. I included the points of the convex hull directly.
```
#include <DGtal/helpers/StdDefs.h>
#include <DGtal/geometry/tools/Hull2DHelpers.h>

#include <algorithm>
#include <vector>

using namespace DGtal;
using namespace functions;

int main()
{
  std::vector<Z2i::RealPoint> hull {
    {804.56832227024199, -68.471176393526704},
    {804.96020257363512, -69.494933490400683},
    {805.35232247974727, -70.519316530915930},
    {825.15497274857830, -114.34711249674335},
    {828.67425867587508, -121.69593677670120},
    {829.05943325037651, -122.39546351563243},
    {829.42436256177677, -122.73833140123513},
    {827.82353937509288, -118.87280445059109},
    {811.06291230576301, -82.124832029926566},
    {806.83483124583609, -72.890457996792406},
    {804.92531970702396, -68.803808853026638},
    {804.55092650722997, -68.125792709291431},
  };

  std::cout << Hull2D::computeHullThickness(hull.begin(), hull.end(), Hull2D::HorizontalVerticalThickness) << std::endl;

  // just by rotating the vector it works fine.
  // so the problem is related to finding the first pair of edges to calculate the resThickness.
  // but this bug misses up the AlphaThickSegmentComputer calculations.

  std::rotate(hull.begin(), hull.begin() + 1, hull.end());
  std::cout << Hull2D::computeHullThickness(hull.begin(), hull.end(), Hull2D::HorizontalVerticalThickness) << std::endl;

  return 0;
}
```

aramadancs:
This is a patch of a fix I was testing locally. I assume it's related to the tolerance of Pi comparison.
[Hull2DHelpers.il.patch.txt](https://github.com/DGtal-team/DGtal/files/7907044/Hull2DHelpers.il.patch.txt)

troussil:
Let me ping @kerautret in my turn! I think he wrote the part about thickness computation.

(Note however that MelkmanConvexHull is guaranteed to properly work only with integer points, but not with real points due to possible inexact float comparisons.)

dcoeurjo:
👍 thx @troussil, I thought it was from you.

Looking at the patch, the angle computation is a bit weird for just a sign of a determinant (even if computed on real numbers). @kerautret ?

[...]

PR #1647, opened 2022-06-19
Fix issue on angle tolerance in computeHullThickness

kerautret:
# PR Description
Fix issue [#1629](https://github.com/DGtal-team/DGtal/issues/1629) on computeHullThickness by adding angle tolerance to detect co-linearity vectors. Thanks @aramadancs!

# Checklist

- [ ] All continuous integration tests pass (Github Actions & appveyor)


src/DGtal/geometry/tools/Hull2DHelpers.ih
```diff
@@ -361,8 +362,8 @@ namespace DGtal
             }
           return 0.0;
           }
-        
-        while(getAngle(aConvexHull[i], aConvexHull[i+1], aConvexHull[j], aConvexHull[j+1]) < M_PI ){
+        while(getAngle(aConvexHull[i], aConvexHull[i+1],
+                       aConvexHull[j], aConvexHull[j+1]) < M_PI - angleTolerance ){
```
dcoeurjo, line 366: why not use a exact geometrical predicate for convex hull computation? If I remember correctly, @troussil has got a collection of 2D predicates already available in DGtal

kerautret: Mainly for the use for alpha-thick segment, If I well remember.

dcoeurjo: Shouldn't we fix this cvx computation with exact computations and not angle comparisons ? 
(for 1.3 or maybe the next one)

kerautret: for non integer input the tolerance is probably useful, else perhaps waiting after 1.3 could let more time

dcoeurjo: ok could you please add an issue on this? Just to remember that this should be fixed.

dcoeurjo (approved):

src/DGtal/geometry/tools/Hull2DHelpers.ih
kerautret: yes I add it

PR #1649, opened 2022-07-08
Add missing <map> header in MeshReader.h files

jeremyfix:
This adds the <map> header in MeshReader.h and MeshReader.ih which appears to be missing although these headers use std::map ;

This was pointed out on my side when compiling the DGTalTools binaries; One of them included MeshReader.h (I think that was not the ih being included) and the compilation failed due to unknown map class.


dcoeurjo:
Thx @jeremyfix for the PR!

dcoeurjo (approved):

PR #1665, opened 2022-10-30
Fix eigen clear() missing method

kerautret:
# PR Description
From using DGtal with ITK that use and transmit its own Eigen version it produces already detected clear() missing method. 
As mentionnes in DGtalTools in issue [354](https://github.com/DGtal-team/DGtalTools/issues/354) of DGtalTools.
It fix both compilation issues of DGtal examples with ITK and the volSurfaceRegularization of DGtalTools.
  
# Checklist

- [x] All continuous integration tests pass (Github Actions & appveyor)


dcoeurjo:
Touching the concept may be a bit dangerous. Are there classes / codes using a clear() on the linear algebra backends ?

kerautret:
> Touching the concept may be a bit dangerous. Are there classes / codes using a clear() on the linear algebra backends ?

yes, mainly in dec with k-forms if I am not wrong. Alternatively perhaps we can extend the type by integrating the  clear() method in order to be conform to CVectorSpace requirements. Trying to avoid ITK transmit its own eigen path should be perhaps not very useful since sooner or later it will produce the same error with newer eigen version.

dcoeurjo:
If it's fine for you, I'll keep this to the release 1.4 or 1.3.1, we need to inspect the sideeffects of that (I suspect cmake tweaks to isolate ITK's Eigen)

kerautret:
yes fine for me, it was just a few edit, I will try the add of clear() alternative perhaps safer but perhaps more on 1.3.1 or 1.4

dcoeurjo:
Actually, I had the same itk issue for #1689 and I also had to fix that (the clear() is not a method of the matrix class anymore apparently).. Thanks for the PR and the bug report, I will close this one to avoid conflicts.

Eigen deps in DGtal is still a bit odd with Eigen behind required by DGTal, and also  installed by polyscope and ITK ..

PR #1709, opened 2023-11-09
Mesh remove isolated vertex

kerautret:
# PR Description

By default the Meshreader imports all vertex even if they are not associated to a face. This PR adds a method to remove them that allows to avoid errors when applying conversion to a SurfaceMesh for instance. It also include a read improvement to ignore path when the material path is wrong and load by default from the current path.

# Checklist


src/DGtal/shapes/Mesh.ih
```diff
@@ -435,14 +437,62 @@ DGtal::Mesh<TPoint>::clearFaces(){
...
+        {
+            face[i]=translateIndexId[face[i]];
+        }
+    }
+}
+
+template<typename TPoint>
+inline
+void
+DGtal::Mesh<TPoint>::changeScale(const typename TPoint::Component aScale){
```
dcoeurjo, line 491: The method name is not great... `rescale` ?

dcoeurjo (changes requested):
Thanks for the PR. Just minor comment

src/DGtal/shapes/Mesh.ih
kerautret: sure ;)

kerautret:
thanks @dcoeurjo just done, but perhaps DGtalTools will failed from the renaming I will do it afterwards

kerautret:
Done small PR in DGtalTools : https://github.com/DGtal-team/DGtalTools/pull/453 that need to be merged before

PR #1712, opened 2023-11-28
Fix Poisson solver in GeodesicInHeat

dcoeurjo:
# PR Description

From https://geometry-central.net:  In theory, it should not be necessary to shift the Laplacian: the Polydec Laplace is always PSD. However, when the  matrix is only positive SEMIdefinite, some solvers may not work (ie Eigen's Cholesky solver doesn't work, but Suitesparse does).

# Checklist


dcoeurjo:
Shouldn't the Poisson problem solver also contain the Dirichlet update of the solver ?

dcoeurjo:
ok selfmerging

PR #1723, opened 2024-05-23
Fix const attribute that shouldn't be

Lieunoir:
# PR Description

Changes a const string attribute to non const, otherwise it conflicts with struct's `operator=` impl

Somehow only gcc 14 gives an error, neither earlier version nor clang complain?

# Checklist


dcoeurjo:
Thanks for the PR. Nope this issue does not seem to be raised for other compilers... interesting..

dcoeurjo:
Is that the only bug you get from gcc14?

Lieunoir:
Yup, works fine otherwise

dcoeurjo:
I'm on the issue you have with testPolygonalCalculus. In the meantime, could you please add a new entry to the Changelog + CONTRIBUTORS files ?

dcoeurjo:
Thanks for the PR, merging

dcoeurjo (approved):

PR #1743, opened 2024-09-19
Upgrading polyscope version 1.2.0→2.3.0

dcoeurjo:
# PR Description

# Checklist


dcoeurjo:
@JacquesOlivierLachaud there is an issue with `tangent_explorer` due to a deprecated API use. Could you please have a look?

I don't know if this is related to this PR but `dgtalCalcuus-halfsphere` only returns a zero value map.

JacquesOlivierLachaud:
Apparently, polyscope SurfaceMesh does not expose any more faces and edges (when I look at surface_mesh.h). So the only way to correct the code is to memorise the vector of faces of the primal surface at construction.

dcoeurjo:
Self-merging this one.

PR #1749, opened 2024-12-05
Adding Boot::multiprecision as a DGtal::BigInteger model

dcoeurjo:
# PR Description

Adding boost::multiprecision for `DGtal::BigInteger`.

# Checklist


dcoeurjo:
Hi @rolanddenis , I may need your help with this PR (cc @JacquesOlivierLachaud )

This PR aims to include `boost::multiprecision` as a (header only) alternative to GMP for mutliprecision integer computations. `DGtal::BigInteger` are now enabled by default, and the backend can be switched to GMP if the user wants that.

I'm fixing a few issues related to WITH_GMP usages and I'm facing a problem in  `PointVector` and the convert you did for explicit casts.

On this PR, if you have a look to https://github.com/DGtal-team/DGtal/blob/4d040fa457b0bca7be38d4450b19a693dcb1d130/tests/kernel/testPointVector.cpp#L136 the `operator-()` is not defined for `DGtal::BigInteger`. It seems to be related to `ArithmeticConversionTraits` you wrote a while ago.

Could you please help us here?

thx

BDoignies:
Hey @dcoeurjo 

I spent some time trying to fix things here but things gets very complicated with Boost.

## Fixing code for GMP

* First, add a specialisation of ArithmeticalTraits for floating-point types 
* Add a way to convert gmp integers to doubles. 
* Wrap the crossProduct operation inside the converting function (see the Boost section, paragraph 2, to understand why only these two are needed). 

This is the easy part. 

## Fixing code for Boost

* Convert any integer to a mixed integer/float operation. This includes every operation in PointVector, but also some more ([BigInt = double - BigInt](https://github.com/DGtal-team/DGtal/blob/cdccf7aa53d9e651fc93db7a5f22e8bdf1313d0b/src/DGtal/geometry/tools/determinant/SimpleIncremental2x2DetComputer.ih#L132)). 

Boost BigIntegers are not meant to be combined with anything other than integers. And the operation listed above barely makes sense to be fair. I am quite surprised it works with GMP to be honest. 

The hardest part is that the tests do not cover everything. For example, only mixed BigInt/float crossProducts are actually checked in PointVector (due to the test you introduced in this PR). So missing any case could result in the error popping up at any time during further development.

## Fixing code for everyone at once

The only stable solution I have is our own BigInteger which is a wrapper around either GMP or Boost depending on the compile options. This allows us to make sure everything works properly and overload the necessary operators and conversions as needed (and hopefully we won't miss any of them :) ).

Your thoughts ? 

PS: To check this PR, do not forget to disable the -WITH_GMP (or DGTAL_WITH_GMP if merged with main2.0) in the github actions, otherwise it won't check against boost.

rolanddenis:
Hello,

sorry @dcoeurjo , I didn't see that mention within my other notifications. Is that the issue you want to mail me about?

I may have time to look at this next week if it still needed regarding the work of @BDoignies on this issue.

BTW, in the line of code you linked to, is the case `double - BigInt` a real (wanted) usage or just an authorized usage since `ResultInteger` is not checked to be a superset of `ArgumentInteger` (or a model of same euclidean ring)?

BDoignies:
Hi @rolanddenis

I do not know if this use is desirable, but it is allowed and  used in (at least) one test ([testInHalPlane-benchmark](https://github.com/DGtal-team/DGtal/blob/cdccf7aa53d9e651fc93db7a5f22e8bdf1313d0b/tests/geometry/tools/determinant/testInHalfPlane-benchmark.cpp#L706)). 

As a side note: there is a bug in the concept check [here](https://github.com/DGtal-team/DGtal/blob/cdccf7aa53d9e651fc93db7a5f22e8bdf1313d0b/src/DGtal/geometry/tools/determinant/SimpleIncremental2x2DetComputer.h#L113C8-L115C74). But the intended meaning is not to check that both types are a model of the same ring.

troussil:
@BDoignies @rolanddenis 
It is quite old, but if I correctly remember, ResultInteger should be a superset of ArgumentInteger. Here, the type "double" is used to hold only integers, which is not what usually means "double". 

(This approach may be interesting because of the following property: a nonzero result guarantees that the computation has been exactly performed; only for a zero result, one has to check if the result is really zero or is zero because of a rounding approximation. This approach turns out to be faster than using big integers when the result is expected to be most of the time nonzero.)
 
Mixing double and big integers is however weird and it is ok for me to not allow this usage.

rolanddenis:
@troussil When you mean that the computation has been exactly performed, you mean that the result is really non-zero even if the computed value may be wrong (due to rounding)?

Anyway, I think that [a line](https://github.com/DGtal-team/DGtal/blob/cdccf7aa53d9e651fc93db7a5f22e8bdf1313d0b/src/DGtal/geometry/tools/determinant/SimpleIncremental2x2DetComputer.ih#L132) like the one pointed by @BDoignies should be fixed. For the context, here is the line :
```C++
myDX = aX - static_cast<ResultInteger>(myX); 
```

We have here the following types:
- `myDX` of type `ResultInteger`,
- `aX` of type `ArgumentInteger`,
- `myX` of type `ResultInteger`,

where `ArgumentInteger` is the first template parameter of `SimpleIncremental2x2DetComputer`, and `ResultInteger` the second one (defaulted to `ArgumentInteger`).

So here, the `static_cast` is useless and the result may be implicitly converted to `ResultInteger` depending on what `ArgumentInteger` really is.

In the [usage example](https://github.com/DGtal-team/DGtal/blob/cdccf7aa53d9e651fc93db7a5f22e8bdf1313d0b/tests/geometry/tools/determinant/testInHalfPlane-benchmark.cpp#L736) cited by @BDoignies , `ArgumentInteger` is `double` and `ResultInteger` is `BigInteger`.
In that case, I think it is reasonable to expect that the result of `aX - static_cast<ResultInteger>(myX)` is a `double` since it is the case if `BigInteger` was replaced by `int`. Thus, the result is implicitly converted back to `BigInteger` and I think it should be explicitly written in the code (does that not triggers a warning from the compiler with some flags?).

So here, I would suggest to replace that line by one of the following possibilities:
1. `myDX = static_cast<ResultInteger>(aX - static_cast<ArgumentInteger>(myX));` (computing in `ArgumentInteger` space)
2. `myDX = static_cast<ResultInteger>(aX) - myX;` (computing in `ResultInteger` space)
3. `myDX = static_cast<ResultInteger>(aX - myX);` (following usual arithmetic conversions) 

I suppose that, following the remark of @troussil , it is the 1st or the 3rd version that is expected here.

hum ok, so I think that at this point, I just arrived at the same level as the first message of @BDoignies :sweat_smile:

rolanddenis:
Some remarks about `Boost.MultiPrecision` (from the [introduction page](https://www.boost.org/doc/libs/1_87_0/libs/multiprecision/doc/html/boost_multiprecision/intro.html)):
- it can wrap the Boost's own arbitrary precision implementation but also the implementation of GMP. So `BigInteger` should always be a `boost::multiprecision::number` whatever the backend that is used,
- I suppose that expression template should be disabled (eg `boost::multiprecision::number<cpp_int_backend<>, et_off>`) so that to get a more predictable result type,
- it seems to support usual arithmetical conversions with fundamental types, so no need to explicitly convert a `BigInteger` to a `double` in the example above,
- to convert a `double` to a `BigInteger`, the `static_cast` is required.

dcoeurjo:
Oh.. that's interesting that boost can wrap several backends (or at least gmp), thx @rolanddenis

BDoignies:
Thanks for the updates @rolanddenis; very interesting and we should use what you pointed out. 

Btw, after a live meeting with @dcoeurjo @troussil and @JacquesOlivierLachaud, we came to the conclusion that the mixing of double and BigInts should be disabled anyway.

rolanddenis:
> So, just about the support of arithmetical conversions, that means that specializing `ArithmeticConversionTraits` should not be needed ...

Maybe needed finally. I need to read the code more carefully...

And we could eventually add implicit conversion to floating-point value for `PointVector` if it is really needed (seems it is tested in `testPointVector`).

tests/kernel/testPointVector.cpp
```diff
@@ -106,6 +106,67 @@ TEST_CASE( "2D Point Vector Unit tests" )
...
+  RealPoint2D p4(t4);
+  
+  Point3D p1_3d( p1[0], p1[1] );
+  Point3D p2_3d( p2[0], p2[1] );
+  RealPoint3D p3_3d( p3[0], p3[1] );
+  RealPoint3D p4_3d( p4[0], p4[1] );
+  
+  SECTION("Cross products with (big)integers")
+  {
+    Point2D p = p1 - p2;
```
rolanddenis, line 136: `p1 + p2` instead?

BDoignies: This is one of the test to remove, it explicitly mix doubles and BigIntegers.

BDoignies: Are you working on the PR right now btw? I just started working on it ^^'

rolanddenis: No, I just tested to modify the code so that all default tests compiles and runs (and with double as added by @dcoeurjo ). It works using essentially the modification you suggest in your first post and some other fixes.

I agree that if we want to make it works in all cases with mixed double/BigInteger, we need to ensure proper conversion in `PointVector` but I don't think it is too much work since there is already some conversion/construction routines in this class. But in  the other hand, there is probably a lot of other things to change in other classes ...

dcoeurjo:
Closing this one. Check #1769 for further details

PR #1771, opened 2025-03-21
Polyscope

BDoignies:
# PR Description

This PR propose a new Viewer based on Polyscope. 

If QLViewer is replaced by this viewer the following changes should be noted:

- Clipping planes are less precise (they do not clip the interior of primitives)
- No more QApplication, show should be called once at the end
- displayMessage now only takes one std::string (not QString)
- Extension work differently: only two callbacks, one for UI / event loop and one for selection.

This PR is incomplete: 
- [ ] Some debug code remaining
- [x] 2D and 3D Images
- [ ] Key descriptions
- [ ] Custom keybinds
- [x] Triangle / Polygone / Prism rendering
- [x] Documentation / ChangeLog, ...


BDoignies:
@dcoeurjo @rolanddenis I have a few question. 

## Regarding images

First, they are not part of the Display3D base class, only of Viewer3D. Should we still maintain them (I think we should)? If so, should the code be moved to Display3D instead?

The use of such images is the second question. Should we render them as textured quads or as voxel grids? 
Right now they are rendered as voxel grid if I am correct, but this means they are not compatible with slice/plane (which can be a problem for 3D images). 

## Regarding keybinds

Which keybinds should be exposed? Polyscope has a lot of functionality that is obsolete with previous keybinds. But some might be interesting?

## Regarding information display

For now, I put the messages inside the user frame of the polyscope. Is this ok? Should they be displayed somewhere else? The same question applies to the keybind description (if any).
 
## Further interface to polyscope 

Currently the viewer has no direct interface to Polyscope. This is a big loss of functionality in my opinion. Polyscope can easily display curvature or normal information for example. Should I add a class that works like this?

```c++
viewer << AddScalarInformation(name, scalar_array_or_iterator);
viewer << AddVectorInformation(name, vector_array_or_iterator);
```

BDoignies:
@dcoeurjo Apart from a minor static analysis issue, is the Polyscope viewer OK for you? Am I waiting for your approval before making all tests and examples use the new viewer.

How should the Polyscope examples be updated (if they should be)?

dcoeurjo:
I have an issue on my clang : 

```
/Users/davidcoeurjolly/Sources/DGtal/src/DGtal/io/Display3D.h:223:30: error: no matching literal operator for call to 'operator""d' with argument of type 'long double' or 'const char *', and no matching literal operator template
  223 |         float voxelWidth = 1.d;
      |                              ^
```

src/DGtal/io/Display3D.h
```diff
@@ -203,9 +207,52 @@ namespace DGtal
...
+     * It can handle both 2D and 3D images, but both 
+     * are rendered as voxel grids. 
+     *
+     * @see Display3D
+     */
+    struct ImageD3D : public CommonD3D {
+        /// Transform
+        Eigen::Affine3d transform;
+        /// Width of each pixel
+        float voxelWidth = 1.d;
```
dcoeurjo, line 223:
```suggestion
        float voxelWidth = 1.f;
```

dcoeurjo:
the `io/viewers` examples are not built when activating the polyscope examples ?

dcoeurjo:
``` 
./testPolyscopeViewer                                                                                                                                                                                                        
[polyscope] Backend: openGL3_glfw -- Loaded openGL version: 4.1 Metal - 89.4
[ERR]VolReader : can't open examples/samples/lobster.vol
libc++abi: terminating due to uncaught exception of type DGtal::IOException: DGtal IO error
[1]    44737 abort      ./
```

dcoeurjo:
@BDoignies As everything is now in #1775 , can we safely close this one?

BDoignies:
Yes, this PR can be closed.

PR #1774, opened 2025-04-24
Ensure that the 3D segment results from the intersection of two maximal 2D segments

copyme:
As in the title. In the previous version we stop extending the 2D segments once we do not have at least two bijective projection, which leads to a problem of the last segment not being maximal. 

# PR Description

*your description here*

# Checklist


copyme:
Seems like there are some issues with the CI (debug mode). Wrong configuration?

dcoeurjo:
The CI bots and make have been considerably improved in the upcoming 2.0 release. One way to fix this would be to close this PR, and create a new one targeting the "main2.0" branch

dcoeurjo:
(and thanks for the PR btw)

copyme:
Hello @dcoeurjo ! Thanks for the reply. I will handle this in the next couple of days.

PR #1776, opened 2025-05-22
Changed Patate to Ponca

BDoignies:
# PR Description

Upgraded Patate to Ponca (fetched by CPM).

I have no problems with Ponca, but the only example that uses it do not compile. Specifically, the error is that it can not convert the result of std::bind, to MetricToPoint in [LocalEstimatorFromSurfelFunctorAdapter:169](https://github.com/BDoignies/DGtal/blob/cdccf7aa53d9e651fc93db7a5f22e8bdf1313d0b/src/DGtal/geometry/surfaces/estimation/LocalEstimatorFromSurfelFunctorAdapter.ih#L169) constructed in [testSphreFitting:104](https://github.com/BDoignies/DGtal/blob/cdccf7aa53d9e651fc93db7a5f22e8bdf1313d0b/tests/geometry/surfaces/testSphereFitting.cpp#L104).

This seems to be related with the fact that the embedder returns a RealPoint while the Metric expect Point (ie. with int). I tried with a simple copy (accepting the loss of precision to test the theory),  but compilation fails later when creating the visitor...  

Any idea @dcoeurjo ?

# Checklist

### DIFF
--- a/src/DGtal/geometry/doc/moduleLocalEstimatorsFromSurfel.dox
+++ b/src/DGtal/geometry/doc/moduleLocalEstimatorsFromSurfel.dox
@@ -86,8 +86,8 @@ several functors:
   the accumulated tensor votes.
 
 - functors::SphereFittingEstimator: this functor requires the
-  WITH_PATATE flag and the Patate Library dependency. The functor
-  performs a sphere fitting on the neighborhood following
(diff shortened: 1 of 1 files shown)


BDoignies:
Maybe, now that the library is automatically fetched, we could add a DGTAL_WITH_PONCA in the CI flags?

dcoeurjo:
Yes please, add Ponca to the CI

nmellado:
Nice !

PR #1782, opened 2025-06-20
Python shortcuts

BDoignies:
# PR Description

Python binding of DGtal (Geometry)Shortcuts. 

This binding relies heavily on the fact that the user would not manipulate the objects outside the Shortcuts framework. Hence, it allows to quickly bind object, without exposing any functions or inside types.

Here is a quick example code:
```python
from dgtal import SH3
from dgtal import helpers
import numpy as np

params = SH3.defaultParameters()
# Probably not the best location for SAMPLES_PATH...
al_capone = SH3.makeBinaryImage(helpers.SAMPLES_PATH + "/Al.100.vol", params)
K = SH3.getKSpace(al_capone, params)
surface = SH3.makeLightDigitalSurface(al_capone, K, params)
surfels = SH3.getSurfelRange(surface, params)
colors = SH3.applyColorMap(np.arange(len(surfels)), params)
SH3.saveOBJ(surface, [], colors, "al.obj")
```
 
Not all shortcuts are bind, especially since there are multiple partial specialization and templated functions in the C++ side. I only bound the ones that are found on the documentation. Hence most of tutorials/shortcuts can now work in Python. Example that are not yet possibles are:
* Count connected surface components. The examples relies heavily on iterating through objects and mappings. I do not know yet how this could be bound and requires some more thought. 
* Saving obj with predicates. Same reason as above, although this seems more straightforward.

Also, I do not know where is the documentation for the python module. Where should I put equivalent of [tutorial-examples/shortcuts.cpp](https://github.com/DGtal-team/DGtal/blob/master/examples/tutorial-examples/shortcuts.cpp) in python? 

# Checklist


dcoeurjo:
Excellent!!

dcoeurjo:
The killing example would be to show results in polyscope (python, not necessarily the viewer) ;)

BDoignies:
Noted; I'll work on this beginning of next week :) !

BDoignies:
Here you are :), any other example you would like to see? Maybe something relating to meshes?

https://github.com/DGtal-team/DGtal/blob/2de77c29184bded4169c3282c2fdc8e228891ded/wrap/tutorials/IntegralInvariantEstimates.py

dcoeurjo:
Changelog entry ?

PR #1842, opened 2026-03-02
Parallel II Estimator

BDoignies:
# PR Description

A new WIP branch for parallel Integral Invariant estimators. 

# Checklist

- [x] I have installed and run the pre-commit hooks (cf [CONTRIBUTING.md](https://github.com/DGtal-team/DGtal/blob/main/CONTRIBUTING.md)).
- [x] New entry in the [ChangeLog.md](https://github.com/DGtal-team/DGtal/blob/main/ChangeLog.md) added.


src/DGtal/geometry/surfaces/estimation/ParallelIIEstimator.ih
```diff
@@ -0,0 +1,150 @@
+#include <ranges>
```
dcoeurjo, line 1: add dgtal header

src/DGtal/geometry/surfaces/estimation/ParallelIIEstimator.h
```diff
@@ -0,0 +1,172 @@
...
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ **/
```
dcoeurjo, line 15: ownership missing ?
@file missing (cf doxygen error)

src/DGtal/geometry/surfaces/estimation/DomainSplitter.h
```diff
@@ -0,0 +1,135 @@
+#pragma once
+
```
dcoeurjo, line 2: gel header here

dcoeurjo (commented):
Looks good to me. Just the doc is missing and some LGPL headers. I've added another domain splitter and a unit test for that.
Code should be in "kernel/domains/" subfolder by the way

dcoeurjo:
I've updated the geometry shortcuts to enable multithreaded curvature estimation.

dcoeurjo:
@JacquesOlivierLachaud would you like to review this PR or I can merge it as is ?

JacquesOlivierLachaud:
Sorry, I forgot. I will have a look at it before Monday.

src/DGtal/geometry/surfaces/estimation/ParallelIIEstimator.h
```diff
@@ -0,0 +1,191 @@
...
+
+    /// @return the grid step.
+    Scalar h() const;
+
+    /**
+     * Attach a shape, defined as a functor spel -> boolean
+     *
+     * @param[in] K the cellular grid space in which the shape is defined.
+     * @param aPointPredicate the shape of interest. The alias can be secured
+     * if a some counted pointer is handed.
```
JacquesOlivierLachaud: "a some" -> "some"

src/DGtal/geometry/surfaces/estimation/ParallelIIEstimator.ih
```diff
@@ -0,0 +1,181 @@
...
+  {
+    myH = h_;
+  }
+
+  template<class TEstimator, typename TSplitter>
+  template<typename It>
+  typename ParallelIIEstimator<TEstimator, TSplitter>::Quantity
+  ParallelIIEstimator<TEstimator, TSplitter>::eval(It it)
+  {
+    // TODO: Initialize properly the estimator (attaching shape, ...)
```
JacquesOlivierLachaud: Remove comment ?

src/DGtal/helpers/ShortcutsGeometry.h
```diff
@@ -1284,12 +1316,40 @@ namespace DGtal
...
+          if ( ! use_parallel )
+            {
+              IINormalEstimator ii_estimator( functor );
+              ii_estimator.attach( K, shape );
+              ii_estimator.setParams( r );
+              ii_estimator.init( h, surfels.begin(), surfels.end() );
+              ii_estimator.eval( surfels.begin(), surfels.end(),
+                                 std::back_inserter( n_estimations ) );
+            }
           const RealVectors n_trivial = getTrivialNormalVectors( K, surfels );
```
JacquesOlivierLachaud, line 1353: A remark here even if it is a bit unrelated to the PR. It would be better if we reorient the normals with the Convolved Trivial estimator (let us say with t-ring=2 or 3). Trivial normal induces problems on many practical digital shapes. (It is a matter of changing Trivial with CTrivial.)

tests/geometry/surfaces/testParallelIntegralInvariantEstimator.cpp
```diff
@@ -0,0 +1,185 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ **/
+
+/**
+ * @file testIntegralInvariantVolumeEstimator.cpp
```
JacquesOlivierLachaud: bad filename

tests/geometry/surfaces/testParallelIntegralInvariantEstimator.cpp
```diff
@@ -0,0 +1,185 @@
...
+  if (rsize != rsizeP)
+  {
+    trace.error() << "Size mismatch between parallel and non-parallel versions: " << rsize << " / " << rsizeP;
+    trace.endBlock();
+    return false;
+  }
+
+  for ( unsigned int i = 0; i < rsize; ++i )
+  {
+    if (std::abs(results[i] - resultsP[i]) >= 1e-2)
```
JacquesOlivierLachaud: Results may be so much different between parallel and sequential (1%) ? Should the measure be relative ?

tests/projects/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ project(SystemWideCompile)
 cmake_minimum_required(VERSION 3.20)
 
 set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_PREFIX_PATH "/home/pattoune/lib/DGtal")
```
JacquesOlivierLachaud: This line seems incorrect and machine-dependent.

JacquesOlivierLachaud (commented):
Just minor comments.

src/DGtal/helpers/ShortcutsGeometry.h
dcoeurjo: I agree... Or at least add an option for that (true by default)

tests/geometry/surfaces/testParallelIntegralInvariantEstimator.cpp
dcoeurjo: That is really strange.. in 3d, the parallel and non-parallel estimators perfectly match (cf testInvarionShortcuts) but not in 2d

dcoeurjo: `Result mismatch between parallel and non-parallel versions at linel 82: 0.0677051 / 0.0676301`

(when I decrease to 1e-6)

JacquesOlivierLachaud: I got the same error when I compile it on my machine. Weird.

JacquesOlivierLachaud: `[ERR]Result mismatch between parallel and non-parallel versions at linel 82: 0.0677051 / 0.0676301`
The error itself is not big, but it should not happen.

JacquesOlivierLachaud: When I asked for one thread only, the parallel version gives the same as the sequential (ouf). When I asked for 8 threads the error is at another linel.
`[ERR]Result mismatch between parallel and non-parallel versions at linel 249: 0.0679939 / 0.0678964`

dcoeurjo:
Yeah.. I wanted to export the shape + domains used in the splitter to better understand the issue... I may suspect boundary issues but that is weird (the same test in 3d is perfectly fine, cf testShortcuts..)

JacquesOlivierLachaud:
What is annoying is that the test is along a digitised disk, so with almost constant curvature. The mismatch cannot be important in this case (all results should be around the constant value). So it may be a true problem. There is also a suspicious used of pointers within a deque that stores the values, but the deque is growing so the pointers may be invalid. Unfortunately, I tried to change this part (without complex views and so-forth), but it does not seem to solve the issue.

